Issue #31, opened 2012-05-12
Request for support on indentation


agladysh:
Hi,

In our company we have relatively strict guidelines on Lua code formatting and several hundreds KLOC of existing code that follows them (so changing guidelines is not a good idea). I wonder, is lua-mode configurable to match our rules, or is it hard to be made flexible enough to be configurable. 

You may find a bunch of code written according to our guidelines, for example, here: https://github.com/lua-nucleo/lua-nucleo/tree/master/lua-nucleo/. I believe that this indentation style is (mostly) not that uncommon.

Indentation rules in short:
- tabs to spaces;
- indent 2 spaces;
- 80 chars max line width;
- multi-line function arguments to be indented with two indents, closing parenthesis — with one (this, probably, is the uncommon part).

As an illustration, I prepared a naïve set of code (most likely incomplete, but will do as a start; if something more significant will be needed, I'll de-indent, say, a bunch of lua-nucleo files, that should cover most of the rules):

https://gist.github.com/37855303239506a16731

There is input, as typed (pasting it in console emacs works for me), expected output, and actual output with current lua-mode Git HEAD.

Unfortunately, we do not have any elisp gurus in-house, so we can't fix the problem ourselves. So I'm humbly asking for assistance. If this is something that is too large to be handled along the "public support" lines, I'll be happy to discuss sponshorship of one kind or another privately, please poke me via GH message.

Please note that we're not asking for a custom implementation of indentation code, but merely for the possibility to configure lua-mode to match our preferences.

Thank you in advance,
Alexander.

P.S. There was an year-old discussion on SO on that matter:

http://stackoverflow.com/questions/4643206/how-to-configure-indentation-in-emacs-lua-mode

I don't speak elisp, so for me that discussion resulted mostly in cargo-cult programming, with little effect.

PR #80, opened 2014-11-07
Rewrite indentation engine

Ambrevar:
The main reason for this change is to change indentation from

```
foobar('arg1', function ()
         print('foobar')
               end)
```

to

```
foobar('arg1', function ()
    print('foobar')
end)
```

This problem is due to alignment. I decided to remove this feature because of the following rationale:
- There is nothing close to a convention on alignment. PIL is not using it.
- Rules on alignment can quickly become ambiguous.
- It can result in ugly indentation. (See above example).
- It is more complex and slower.
- When indenting with tabs, one needs to use Emacs' smart-tabs plugin (to follow the "indent with tabs, align with spaces" rule). Not many editors can do that.

This rewrite partly addresses issue #31 and some issues in indentation-test.el.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -1316,19 +1316,18 @@ return it."
   "Return appropriate indentation for current line as Lua code."
   ;; Algorithm: indentation is computed from current line and previous line.
   ;; Current line:
-  ;; - one level on every unmatched closer if first token is a closer
-  ;; - one level if first token is 'middle
+  ;; -1 on every unmatched closer if first token is a closer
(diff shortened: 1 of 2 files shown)


immerrr:
This is a long-standing PR, and given that this is already supported with `lua-indent-nested-block-content-align` and `lua-indent-close-paren-align` variables, I am inclined to respectfully close it.

Please, let me know if you have a reason to believe I need to rethink this decision.

Ambrevar:
No problem, I haven't use Lua in years... :p

PR #82, opened 2015-02-22
Make browser customizable within the mode

holomorph:
I generally find it nice when I don't have to touch browse-url custom variables, and can set the browser just for modes that can search in html documentation, since I find eww to be really nice for simple html docs.

Also, try to use local documentation.

I think it would be nicer to have a list of common locations for the lua documentation in various distros and then try to find a suitable local doc from that list, but the way I'd want to do it is `(seq-some-p #'file-exists-p list-of-places)`.  If you're fine with depending on the seq library, which is in future Emacs (otherwise GNU ELPA), then I'll try doing that.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -222,8 +222,19 @@ Should be a list of strings."
   :type 'boolean
   :group 'lua)
 
-(defcustom lua-search-url-prefix "http://www.lua.org/manual/5.1/manual.html#pdf-"
-  "*URL at which to search for documentation on a word"
+(defcustom lua-documentation-function 'browse-url
+  "Function used to fetch the Lua reference manual."
+  :type `(radio (function-item browse-url)
+                ,@(when (fboundp 'eww) '((function-item eww)))
+                ,@(when (fboundp 'w3m-browse-url) '((function-item w3m-browse-url)))
+                (function :tag "Other function"))
+  :group 'lua)
+
+(defcustom lua-documentation-url
+  (or (and (file-readable-p "/usr/share/doc/lua/manual.html")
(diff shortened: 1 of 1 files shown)


immerrr:
Thank you for the contribution, the idea looks interesting.

The name `lua-browse-url-function` hints that this variable overrides `browse-url` for `lua-mode` which is not true and, if anything, is easier to do with `make-variable-buffer-local`. That being said, having configurable function for documentation browsing only is a nice addition.

`lua-browse-url` variable name should also somehow reflect that it's about documentation rather than generic browsing.

Also, I'd like to know if you have a [copyright assignment for FSF](https://www.gnu.org/licenses/why-assign.en.html) or would you be willing to sign one? I still hope that one day lua-mode will finally get into Emacs trunk and your contribution is borderline *significant* (significant starts from 15 LOC).

holomorph:
Thanks for the feedback, I'll try to make the appropriate changes.

I just recently submitted copyright assignment forms for FSF, so I should be set rather soon.

immerrr:
Also, it feels nice to have some sane defaults spelled out explicitly:
```lisp
  :type `(radio (function-item browse-url)
                ,@(when (fboundp 'eww) '((function-item eww)))
                (function :tag "other function"))

```

holomorph:
Whoa, that's fancy.  I added w3m to the list of defaults.

immerrr:
> (I'm not sure if I should squash these or let you do it)

Feel free :)

holomorph:
Squashed!

immerrr:
Merged, thank you!

PR #83, opened 2015-02-22
Tell lua-mode about prettify-symbols-mode

holomorph:
Given the treatment of functions in lua, I feel like this makes sense :dancer:


immerrr:
I tend to disagree with putting this into lua-mode:
- this makes non-anonymous function definitions ugly
- this change is very self-contained and requires no interaction with lua-mode
- Lua is not really a functional language, despite having anonymous functions and TCO

PR #88, opened 2015-03-24
Add file test/.nosearch

tarsius:
This prevents that "test/" is added to the `load-path` when using
tools such as built-in `normal-top-level-add-subdirs-to-load-path`.


immerrr:
Merged, thank you. And thank you for your work on magit.

tarsius:
Thanks and you're welcome :-)

ramnes:
The `.nosearch` file should be in the `test/` directory, not at the root of the repository.

My config is currently broken because of that ("File error: Cannot open load file, no such file or directory, lua-mode").

PR #91, opened 2015-06-10
Add # length operator to syntax as punctuation

bstpierre:
Using the `#` length operator was being highlighted as a comment, and bleeding through all the code below that point.

```
a = {1,2,3}
print(#a)
print("this should not be highlighted as a comment")
```

With point at "this", `describe-face` was reporting "font-lock-comment-face". After this patch, it correctly reports "font-lock-string-face".

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -683,6 +683,7 @@ Groups 6-9 can be used in any of argument regexps."
     (modify-syntax-entry ?\" "\"")
 
     ;; single-character binary operators: punctuation
+    (modify-syntax-entry ?# ".")
     (modify-syntax-entry ?+ ".")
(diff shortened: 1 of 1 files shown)


immerrr:
Hi and thank you for the report.

`#` should not be highlighted as a comment, at least it does not on clean-slate emacs23, emacs24 and emacs25 I have at my disposal.  Another thing is that it shouldn't "bleed through", because the syntax table explicitly marks newline as `endcomment`.

Could you check that it still does that when you do `emacs -Q -l lua-mode.el`? What system and emacs do you use?

immerrr:
The issue described in this PR should not happen in clean Emacs, and there was no evidence of it happening provided, so I'm closing this PR.

PR #92, opened 2015-06-12
Move .nosearch in test/

ramnes:
See https://github.com/immerrr/lua-mode/pull/88#issuecomment-111476708

PR #93, opened 2015-06-12
correctly place .nosearch in subdirectory

tarsius:
Re #88

Sorry about that.


tarsius:
Never mind, I did not scroll enough to see #92.

PR #97, opened 2015-07-07
Add `run-lua` alias

marsam:
Hi:
I've added the `autoload` cookie to the `lua-start-process` function and added an alias `run-lua` to it, because is common to use the `run-` prefix for repls


immerrr:
Merged, thank you.

PR #105, opened 2015-09-14
Add lua-hanging-indent-level.

melisgl:
Examples with lua-hanging-indent-level and lua-indent-level being 4
and 2 respectively:

local foo = functionName(
    varOne, varTwo, varThree, varFour)

local foo = functionName(varOne, varTwo,
                         varThree, varFour)

local foo = {
    key1 = value1
    key2 = value2
}

local foo = { longTableKey = value1
              anotherTableKey = value3
}

something =
    somethingElse

if foo then
  doSomething
end


lua-mode.el
```diff
@@ -230,6 +230,15 @@ for Emacsen that doesn't contain one (pre-23.3)."
   :type 'integer
   :group 'lua)
 
+(defcustom lua-hanging-indent-level 3
```
immerrr: Any reason why this is not called `lua-continuation-indent-level`?

melisgl: On Wed, Sep 23, 2015 at 9:40 AM, dennis again <notifications@github.com> wrote:
>
> In lua-mode.el:
>
> > @@ -230,6 +230,15 @@ for Emacsen that doesn't contain one (pre-23.3)."
> >    :type 'integer
> >    :group 'lua)
> >
> > +(defcustom lua-hanging-indent-level 3
>
> Any reason why this is not called lua-continuation-indent-level?


I used a different word because not all line continuations are
indented this way. In particular:

    local foo = functionName(varOne, varTwo,
                             varThree, varFour)

Feel free to change it, if you prefer something else.

Cheers,
Gabor

PR #112, opened 2015-12-10
moved ruby's electric indent to lua-mode

fangzhouuu:
Copied from ruby-mode.el

now, when typing a closing token like _end_, an automatically reindent
will be performed. it also works for _else_  _elseif_  _until_. I did a quick
scan on lua syntax, described in lua manual and don't think there will be
 more keyword which trigger a reindent.

example:
```
1 function foo()              1 function foo()
2     return "foo"            2    return "foo"
3     en                      3 end
        ^ type 'd' here will cause ^ to reindent

```

I'm been using this for almost a month and everything seems to be ok, though in fact I'm not very familiar with elisp, so better have a look at the code and do some extra test.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -436,6 +436,12 @@ traceback location."
     ["Search Documentation" lua-search-documentation t])
   "Emacs menu for Lua mode.")
 
+
+;; auto indent after typing keywords below
+(defconst lua-block-mid-keywords
+  '("else" "elseif" "until")
+  "Keywords where the indentation gets shallower in middle of block statements.")
+
 ;; the whole defconst is inside eval-when-compile, because it's later referenced
 ;; inside another eval-and-compile block
 (eval-and-compile
@@ -748,6 +754,9 @@ Groups 6-9 can be used in any of argument regexps."
   (with-no-warnings
     (lua--setq-local comment-use-global-state     t))
   (lua--setq-local imenu-generic-expression       lua-imenu-generic-expression)
+
+  (add-hook 'electric-indent-functions 'lua--electric-indent-p nil 'local)
+
   (when (boundp 'electric-indent-chars)
     ;; If electric-indent-chars is not defined, electric indentation is done
     ;; via `lua-mode-map'.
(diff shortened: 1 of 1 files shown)


immerrr:
Hi and thank you for the PR.

Electric indentation of block closing tokens is already implemented via `abbrev-mode`, see [lua-mode-abbrev-table](https://github.com/immerrr/lua-mode/blob/0b41d44cd1f34af5f74052cea50135c590bdd30e/lua-mode.el#L678-L690) as well as [lua-electric-flag variable](https://github.com/immerrr/lua-mode/blob/0b41d44cd1f34af5f74052cea50135c590bdd30e/lua-mode.el#L343-L346). That being said, there were plans of migrating to electric-indent infrastructure, but unfortunately that is only available since emacs-24.4 and some still use older versions.

If you don't mind doing some refactoring to select the one that is appropriate for the running version of emacs on top of your slightly adapted copy-n-paste, I'll gladly accept that.

fangzhouuu:
Thanks for you quick reply.

I'm using emacs 24.5 actually, so I guess there may be some surprise for me since electric indent doesn't work for me out of the box(I've tried toggle =lua-toggle-electric-state= but didn't see any difference)

And as for the code I copied, I'll try to make them clean to provide those who need a fallback. Got to learn something but that will be OK, just need some time.

immerrr:
@fangzhouuu you might need to enable `abbrev-mode`, it got disabled by default in one of the recent releases

fangzhouuu:
Well I've enable abbrev-mode but it still not working for <code>end else elseif</code>(works for brackets). It could be the the effect of other package since I copied a huge bunch of code in my config file. I might have to setup a clean emacs dev envirnoment using a VM or a docker container.

immerrr:
You can do `emacs -Q -l path/to/lua-mode.el` to try out a fresh emacs env.

If you have [cask](https://github.com/cask/cask), you can use `make tryout` command from `lua-mode` directory.

immerrr:
Also, there are tests for electric indentation functionality: https://github.com/immerrr/lua-mode/blob/0b41d44cd1f34af5f74052cea50135c590bdd30e/test/test-electric-mode.el#L35-L71

PR #115, opened 2016-04-29
Escape tabs in lua-make-lua-string

RobertCochran:
Doing so prevents comint from intercepting the tabs and passing them to
the underlying shell process, which inserts shell tab completion text
into the code string, mangling it so that it can't run.

-----

As noted in the commit description. This probably isn't the right solution; the
better one would be to tell comint not to touch the input string, but I'm not
sure how to do that. Nonetheless, it's still a solution, and it fixes the bug at
hand.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -1708,10 +1708,14 @@ This function just searches for a `end' at the beginning of a line."
     (with-temp-buffer
       (insert str)
       (goto-char (point-min))
-      (while (re-search-forward "[\"'\\\n]" nil t)
-        (if (string= (match-string 0) "\n")
-            (replace-match "\\\\n")
-          (replace-match "\\\\\\&" t)))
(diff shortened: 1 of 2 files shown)


RobertCochran:
I noticed that the CI build has failed. Reading the logs, it's not entirely
clear what error (if any) has occured. The `make` error is merely
`Opening directory: Not a directory, /home/travis/build/immerrr/lua-mode/lua-mode.25.elc`.
Have I done something incorrectly/failed to do something correctly?

immerrr:
It appears to be a backward-incompatible change in buttercup which i have proposed to fix in https://github.com/jorgenschaefer/emacs-buttercup/pull/59.

immerrr:
The build passes now, could you add a test for the new behaviour?

RobertCochran:
Test added. Please let me know if you need me to do anything else.

immerrr:
Thank you!

PR #119, opened 2016-06-06
Support querying subprocess for completion.

### LINKED ISSUE #87: company backend
It be great if lua-mode came with a company backend.

technomancy:
Fixes #87.

I ended up disabling the feature that pulls completion targets from locals by default since it can cause unexpected side-effects, but it's documented in the readme how to turn it on.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -1929,6 +1929,7 @@ This is distinct from `backward-sexp' which treats . and : as a separator."
 Queries current lua subprocess for possible completions."
   (let* ((start-of-expr (lua-start-of-expr))
          (expr (buffer-substring-no-properties start-of-expr (point)))
+         (expr (car (last (split-string expr "\n")))) ; avoid multi-line input
          (libs (lua-local-libs))
          (file (make-temp-file "lua-completions-")))
     (lua-send-string (lua-completion-string-for expr libs file))
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -769,6 +769,9 @@ Groups 6-9 can be used in any of argument regexps."
       (setq mode-popup-menu
             (cons (concat mode-name " Mode Commands") lua-emacs-menu)))
 
+  (make-variable-buffer-local 'completion-at-point-functions)
+  (add-to-list 'completion-at-point-functions 'lua-complete-function)
+
   ;; hideshow setup
   (unless (assq 'lua-mode hs-special-modes-alist)
     (add-to-list 'hs-special-modes-alist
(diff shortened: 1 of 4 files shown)


technomancy:
Another note is that this *almost* works for completion in the lua subprocess buffer, except evaluating the string to determine the completions causes the prompt to be re-printed, which throws off the function which inserts the completions. If there were a way to suppress the re-printing of the prompt, it would solve that problem.

immerrr:
Hi and thank you for a PR for a long waited feature.

lua-mode.el
```diff
@@ -769,6 +769,9 @@ Groups 6-9 can be used in any of argument regexps."
       (setq mode-popup-menu
             (cons (concat mode-name " Mode Commands") lua-emacs-menu)))
 
+  (make-variable-buffer-local 'completion-at-point-functions)
```
immerrr: For making other people's variables local just in this mode it's better to use `make-local-variable`.

lua-mode.el
```diff
@@ -1814,6 +1817,105 @@ If `lua-process' is nil or dead, start a new process first."
...
+  "Find all modules loaded with require which are stored in locals.
+
+Returns a list of lists where the car is the local name and the cadr
+is the string that is passed to require."
+  (save-excursion
+    (when lua-local-require-completions
+      (let ((libs nil))
+        (goto-char (point-min))
+        ;; find each match of "local x = require y" and save for later
+        (while (search-forward-regexp lua-local-require-regexp nil t)
```
immerrr: It's better to use `lua-find-regexp` here, as it will skip expressions found in strings and comments.

lua-mode.el
```diff
@@ -1814,6 +1817,105 @@ If `lua-process' is nil or dead, start a new process first."
...
+         (expr (buffer-substring-no-properties (1+ start-of-expr) (point)))
+         (libs (lua-local-libs))
+         (file (make-temp-file "lua-completions-")))
+    (lua-send-string (lua-completion-string-for expr libs file))
+    (sit-for 0.1)
+    (list start-of-token (point)
+          (when (file-exists-p file)
+            (with-temp-buffer
+              (insert-file-contents file)
+              (delete-file file)
```
immerrr: Is it intentional that we don't delete the file if something goes south? Should we wrap it into `unwind-protect`?

immerrr:
I have been meaning to integrate Jupyter for REPL purposes for ages, but never had time for it. Your solution has the undeniable benefit of not having any extra dependencies and may work for now. One thing though is that I'm not very comfortable accepting it without any tests. Could you add some?

lua-mode.el
```diff
@@ -1814,6 +1817,105 @@ If `lua-process' is nil or dead, start a new process first."
...
+          (add-to-list 'libs (list (match-string-no-properties 1)
+                                   (match-string-no-properties 2))))
+        libs))))
+
+(defun lua-complete-function ()
+  "Completion function for `completion-at-point-functions'.
+
+Queries current lua subprocess for possible completions."
+  (let* ((start-of-expr (save-excursion
+                          (search-backward-regexp "[^\.a-zA-Z0-9_]") (point)))
```
immerrr: Does it handle `:` as binding attribute reference (or whatever is it actually called in Lua)? Can it?

Same questions for whitespace between components? E.g.
```
my_super_long_variable_name.
    super_long_attribute_name
```

immerrr:
> I've made all the changes you suggested, though the last one was pretty tricky!

Great, thank you!

> Emacs cannot safely make TLS connections

Woah, thank you for the heads up, never came up to me that Emacs is that careless.

> Do you have any recommendations for how that would work?

There's `package-install-file` function that could be run on a local `tar` archive, will that work for you?

technomancy:
> There's package-install-file function that could be run on a local tar
> archive, will that work for you?

Oh, sorry that wasn't clear--I am able to install buttercup just fine;
it's just that I can't figure out how to run the tests since the
Makefile assumes Cask is installed, and I don't want to install Cask
since it looks very insecure.

Is there a way I can run the tests independently? The only interactive
function from buttercup I see is buttercup-run-at-point, but I want to
run all the tests.

immerrr:
Something like this should work:
```
emacs -batch -l lua-mode.el -l buttercup -f buttercup-run-discover
```

technomancy:
Oh, of course. I was looking for interactive functions; should have
known it would run from the shell instead.

I added some tests, both for the standard completion and the one that
scans for local requires.

immerrr:
I'll do my best to find time to give it a spin on the weekend and merge, thank you for your effort.

immerrr:
Hi!

Apologies for my replies that take eternity, I rebased it, tried it around a bit, done a couple of changes in https://github.com/immerrr/lua-mode/tree/pr-119, feel free to reset your head on top of it or use it as a reference. Aside from that, it looks great, thank you! 

Have you signed an FSF copyright assignment by chance?

immerrr:
Whoops, I seem to have broken something with my changes...

immerrr:
> I am seeing breakage, but it's not related to your change.

I was trying to replace `replace-regexp` (which generates a warning for being intended to be used interactively) with `replace-regexp-in-string`, but tests started failing afterwards. I'll have a look at it on Saturday if you don't beat me to it.

> I have signed copyright paperwork; is lua-mode going to be integrated into Emacs? That's great news.

That was the intention a while ago. There's still a big blocker in the shape of the indentation engine, that comes from the ages when no one cared about copyright assignments, but I'm trying to at least avoid creating more.

Although the way Emacs development goes right now, in all likelihood the only win would be putting `lua-mode` to `elpa` instead of (or in addition to) `melpa`.

> I realized that this code could be improved a fair bit by including top-level locals in the completion targets; would you like me to add that on your branch along with the bug fix?

That would be nice to have. And you don't have to rush it to get to this PR.

immerrr:
CI still seems to fail, do you see how to fix it?

immerrr:
@jdtsmith I am all for it, but unfortunately don't have time to debug this.

jdtsmith:
I found a few issues with @technomancy's PR and generated a new corrected PR #147.

technomancy:
Looks like this one is no longer needed; thanks for taking it over!

Issue #89, opened 2015-04-11
Eval error


tofferPika:
I'm falling into a strange behavior. The following chunk: https://gist.github.com/tofferPika/4c4cc8e004a38137e163 run fine from the terminal with the lua interpreter (5.2) but refuse with the lua-mode process. I can't figure where the code break the string passed to the eval process, I first thinked it was the nested varargs, or some function name identical to lisp one... Additionaly, when moving functions arround and add blank lines here and there, the error message change.
I hope it make sense

immerrr:
Thank you for the report, however, I can't reproduce the issue:
![screenshot from 2015-04-12 08 21 39](https://cloud.githubusercontent.com/assets/579798/7104595/02e0c944-e0ed-11e4-807a-b421dbe7a9d4.png)

Could you fetch the current master and reproduce the issue using `emacs -Q -l path/to/checkout/lua-mode.el`?

tofferPika:
I've this issue from the current master. Also I forgot to mention it is from OS X Emacs 24.4 (0.9). I've the same result when running with the -Q -l flags. There is a screenshot here : https://dl.dropboxusercontent.com/u/50413630/lua-mode-shot.png. Tell me how can I  help you to find the problem, I'm not a "Lisper" but I should be  able to read and understand the logic.

immerrr:
Here's where you have the complete command that is being sent to the subprocess and that has to be syntactically-correct Lua code:
https://github.com/immerrr/lua-mode/tree/master/lua-mode.el#L1789

You can insert `(message "%s" command)` at that line and check the output in the `*Messages*` buffer.  I'd guess that something goes wrong in `lua-make-lua-string` function that should turn the source code being sent into a properly escaped Lua string. If you don't see anything obvious, put the string from `*Messages*` buffer here and I'll try to have a look at it.

tofferPika:
Ok, after some tests, I found it is not a string escaping problem. It happen when there is more than 500 characters sent. I've checked process-send-string doc, and effectively it mention that if the string is more 500 characters long, it is sent in several bunches. But I stick with why only the first part is evaluated before the complete string is sent.

technomancy:
My workaround for this is to use my own function instead of `lua-send-buffer` which hits this problem:

```lisp
(defun pnh-lua-send-file ()
  (interactive)
  (lua-send-string (format "_ = dofile('%s') print('ok')" buffer-file-name)))
```

[...]

PR #121, opened 2016-06-22
Fix sending of long code strings to REPL using temporary files (#89) 

arbv:


arbv:
It seems there are problems with using temporary files on Travis. I will try to solve this issue in the different way.

PR #122, opened 2016-06-22
Fix sending of long code strings to REPL using temporary files (#89)

arbv:
Now we will try to handle long chunk strings (> 500 characters) via temporary files.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -1779,12 +1779,53 @@ When called interactively, switch to the process buffer."
   (set-marker lua-region-end (or arg (point))))
 
 (defun lua-send-string (str)
-  "Send STR plus a newline to Lua subprocess.
+  "Load STR plus a newline into Lua subprocess.
 
 If `lua-process' is nil or dead, start a new process first."
-  (unless (string-equal (substring str -1) "\n")
-    (setq str (concat str "\n")))
-  (process-send-string (lua-get-create-process) str))
+  (let ((tmp-file (make-temp-file "luamode" nil ".tmp"))
+        (lua-process (lua-get-create-process)))
+    ;; write data into temporary file
+    (with-temp-buffer
+      (insert (if (string-equal (substring str -1) "\n")
+                  str
+                (concat str "\n")))
+      (write-file tmp-file))
+    ;; evaluate data in the temporary file and then remove it
+    (process-send-string
+     lua-process
+     (format (concat
+              "\n"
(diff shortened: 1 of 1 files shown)


jtecca:
Is there anything that can be done (besides fixing the merge conflicts) to get this PR merged?  I am running into issues with long buffers and would like to see this merged into the main repo.

I am willing to help if anything is needed.

immerrr:
Hi, please accept my apologies, this PR somehow flew under the radar for me.

I think this feature has to be off by default, because sending via tempfiles won't work over network. Alternatively, it could be on by default for macOS, and off by default otherwise.

Also, could the Lua part be simplified a bit? 
- only protect the part up to parsing the code by pcall, return a callable chunk from that, and then remove the file itself, and then report the error if it failed or execute the chunk if it didn't
- use `assert` on `io.open` to do the same with less code
- same for `f:read` although I don't think it returns an error message (does it?)

arbv:
I believe this feature should be on by default on Windows too. I made this PR initially because I had problems on Windows.

immerrr:
Oh, then I was wrongly under impression that it was about macOS! Then yes, it should be on by default for Windows.

vacech:
I just wanted to leave a short note based on my experience with  Aquamacs and the temporary file code (as posted in [9eddee](/arbv/lua-mode/commit/9eddee17bc2ef1b9118449c1d750e1ef143d5786)). 
The name of every temporary file written by `lua-send-string` is added to the Open Recent menu which soon makes it cluttered with .tmp names. For my usage I modified the code to use the same file repeatedly for the whole Aquamacs session. The other way around is to prevent the temporary file names completely from going into the recent file menu by adding a matching regexp to  `recentf-exclude` (`".*/luamode.*\.tmp"` should do).

vacech:
You could also consider simplifying the part of `lua-send-string` that sends the command to process the temp file to
```lisp
(process-send-string
     lua-process
     (format "local tmp = '%s'; local _, e = pcall(function () dofile(tmp) end);
 if e then error(e) end; os.remove(tmp)\n" tmp-file))
``` 

which for me seems to work well and inserts way less text into the lua REPL.

aabacchus:
RE: what platforms is this relevant to, the default LUA_MAXINPUT is 512 for all platforms: https://www.lua.org/source/5.4/lua.c.html#LUA_MAXINPUT

PR #131, opened 2016-12-28
Wrap lua--fill-paragraph in save-excursion

jorams:
The internals of the function should not change the location of point.
This also matches how the native fill-paragraph functionality works.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -810,12 +810,13 @@ Groups 6-9 can be used in any of argument regexps."
   ;; does not find a comment at the beginning of the empty line before the
   ;; comment and falls back to text-based filling ignoring comment-start and
   ;; spilling the comment into the code.
-  (while (and (not (eobp))
-              (progn (move-to-left-margin)
-                     (looking-at paragraph-separate)))
-    (forward-line 1))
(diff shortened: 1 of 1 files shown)


immerrr:
Hi, and thank you for the contribution. I'll try to take a look at it as soon as I can.

I remember there were some issues with saving point location when stuff got changed around it (like, in indentation function the code safes offset from the end rather than offset from the beginning). Could you verify that save-excursion works ok when fill-paragraph shuffles words & whitespace around?

jorams:
I have been using this code for a few days now, and I've only found one case where the behaviour is slightly imperfect, but still as good as it is before this change.

That situation is as follows:

- You have a comment line that starts with `--`
- Point is somewhere before the first non-whitespace character on that comment line
- You call `fill-paragraph` and it changes something on that line.

Then point will move to the first non-whitespace character after the `--`, instead of staying exactly where it's supposed to be.

Interestingly, when you add another `-` this imperfection does not occur.

In all other cases I've encountered, the behaviour was exactly as desired. Point maintains its position in the text, even if that text moves around a lot.

immerrr:
Great, thank you.

PR #141, opened 2018-01-04
Assume Emacs >= 24.3

rrthomas:


lua-mode.el
```diff
@@ -1522,25 +1447,19 @@ one."
...
-         (seq (eval lua--function-name-rx) (* blank) (regexp "\\(?1:\\)[{(]"))
-         (seq (or
-               ;; assignment statement prefix
-               (seq (* nonl) (not (any "<=>~")) "=" (* blank))
-               ;; return statement prefix
-               (seq word-start "return" word-end (* blank)))
-              (regexp "\\(?1:\\)")
+     (or (seq (group-n 1 symbol-start "local" (+ blank)) "function" symbol-end)
+
+         (seq (group-n 1 (eval lua--function-name-rx) (* blank)) (any "{("))
```
immerrr: I think `group-n 1` is misplaced here. Admiddetly, it's a strange application of group-n and my memory is hazy on the details, but the way it's used in `lua-point-is-after-left-shifter-p` suggests that the end of `?1:` group should denote the beginning of the open-block token. You've moved the `?1:`'s around in a different manner and it makes me wonder why did the tests not fail.

rrthomas: I'm sorry, I am not very familiar with `rx`, and I don't have a good memory of what I learnt to make this patch.

I've just studied the code again. I tried moving the `group-n 1` into sexps by themselves, corresponding to their previous positions, but this broke the tests.

Reading your description again, I think the current placement does indeed correspond with what you want: the end of each `group-n 1` is indeed the end of the token directly before the open-block token. The difference from before is that the groups now match something before it as well.

immerrr: Ok, yes, it makes sense now, thank you

PR #148, opened 2019-05-24
Completion caching (simplified)

jdtsmith:
This supersedes #147.   It is simply a cleaned-up and much-simplified history of all the cumulative changes from #119 and the later additions of breadth-first globals tree traversal, caching, and comint-redirect interface with the lua process begun in #147.  

I did not document the ability to make `lua-process-buffer` buffer-local, which might be a nice addition.


emacs_lua_complete.lua
```diff
@@ -0,0 +1,81 @@
...
+--     complete as well
+--
+-- Example:
+-- __emacs_lua_complete({'table','ins'},{{var="xyz",lib="libfile"}},
+--                      {"frank","beans"},5000)
+--
+-- Written 2019/05 J.D. Smith
+function __emacs_lua_complete(parts,libs,locals,limit)
+   -- Setup a FIFO queue for breadth-first tree traversal
+   local queue={head=0,tail=0}
```
immerrr: Coming from Python I feel like this line is missing whitespaces around equals signs (or binary operators). What are the common style guidelines in Lua world? I saw some packages defining whitespace requirements, and some just deferring to "whatever is the default in PIL".

emacs_lua_complete.lua
```diff
@@ -0,0 +1,81 @@
+--    __emacs_lua_complete:
```
immerrr: Nice step to move this away, it is clearly way too complicated to include in `lua-mode.el` file verbatim. One extra thing to do here is to figure out the packaging part. For example, we could put this into `scripts/` directory, and update MELPA recipe and the Cask file to include `(:files :defaults ("scripts", "*"))`.

emacs_lua_complete.lua
```diff
@@ -0,0 +1,81 @@
...
+	 globals[v.var]=good and lib or {}
+      end
+   end
+   if locals then
+      for _,v in ipairs(locals) do globals[v]={} end
+   end
+
+   -- Descend the tree to the limit of dotted parts
+   local g,pre=globals,""
+   for i=1,#parts do
```
immerrr: Would `ipairs` work here to replace all the `parts[i]` in the body of the loop?

emacs_lua_complete.lua
```diff
@@ -0,0 +1,81 @@
...
+      if not g then break end	-- should always exist
+      if i<#parts then		-- keep descending
+	 pre = (i > 1 and pre.."." or "")..parts[i]
+	 g = g[parts[i]]
+      else  -- final part; seed the queue with any matching tables
+	 for k,v in pairs(g) do
+	    if type(k)=="string" and k:sub(1,1) ~= "_" and k:find('^'..parts[i]) then
+	       local dpath=(i > 1 and pre.."." or "")..k
+	       print(dpath)
```
immerrr: Would it be better to refactor this function to return a list, so that it is the responsibility of the outer code to decide on how to report the list to the parent process, be it `print`, or `io.write`, or json serialization or whatever?

immerrr: Or maybe make it an iterator function somehow (I haven't written those in Lua for ages), to keep returning more and more values, and then have the wrapping function apply the `limit` and serialize the elements accordingly.

lua-mode.el
```diff
@@ -86,12 +86,11 @@
 
 ;;; Code:
 (eval-when-compile
-  (require 'cl))
-
+  (require 'cl)
+  (require 'subr-x))
```
immerrr: We should probably add `(emacs "24.4")` to package dependencies in `Cask` (https://github.com/purcell/package-lint/issues/110 for reference)

lua-mode.el
```diff
@@ -304,7 +315,7 @@ If the latter is nil, the keymap translates into `lua-mode-map' verbatim.")
  key like `{' is pressed")
 (make-variable-buffer-local 'lua-electric-flag)
 
-(defcustom lua-prompt-regexp "[^\n]*\\(>[\t ]+\\)+$"
+(defcustom lua-prompt-regexp "^\\(>[\t ]+\\)+"
```
immerrr: Will this regexp break for code that does `io.write` and doesn't append a newline? E.g.
```
$ lua
Lua 5.2.4  Copyright (C) 1994-2015 Lua.org, PUC-Rio
> io.write('asdf')
asdf> 
```

lua-mode.el
```diff
@@ -1623,10 +1634,17 @@ This function just searches for a `end' at the beginning of a line."
...
+  (expand-file-name "emacs_lua_complete.lua"
+		    (file-name-directory load-file-name))
+  "Path to the associate lua completion function definition.")
 (defvar lua-process-init-code
   (mapconcat
    'identity
-   '("local loadstring = loadstring or load"
+   `(
+     ,(concat "dofile(\"" lua-process-complete-code-filename "\")")
+     "os.execute('stty -echo -onlcr')"
```
immerrr: Will this return an error if there is no `stty` on the host? E.g. on windows.

immerrr:
Hopefully, one last round of questions. I am sorry it takes forever, but it is just that every rewrite or functionality change has the potential of generating more questions and doubts. Smaller changes are much easier and faster to get in.

emacs_lua_complete.lua
jdtsmith: lua is very lenient about whitespace in either direction, but I can certainly add it in for visual clarity. Added.  BTW, I will alter history in this branch after the fact to avoid the clutter of small commits like these, so please don't rely on specific commit ID's!

emacs_lua_complete.lua
jdtsmith: Yeah.  It also saves constant recompilation every time you complete, and opens the door for more "behind-the-scenes" interaction with lua going forward.  

The scripts sounds like a good plan. I created a `scripts` directory and referenced it.  I'm not really familiar with Cask; I think you have push access to this branch, if you don't mind sending a commit over.

emacs_lua_complete.lua
jdtsmith: Well, I use `i` to tell me if I'm in the "last" part, which may be a partial match (or even an empty string for `table.[M-Tab]` type completions).  But I'm by no means a lua expert, so let me know if there's a better lua way.

emacs_lua_complete.lua
jdtsmith: It's a pretty small loop, often only one element, so an iterator seems a bit overkill to me.

lua-mode.el
jdtsmith: Right.  Again I'm not a Cask expert; would you mind adding the correct incantations and submitting to to this PR branch?  I'll stop rewriting pushed history until we converge.  Anything that needs to be placed elsewhere just add a `[TOSQUASH]`.

lua-mode.el
jdtsmith: I think it will just silently do nothing.

```lua
> os.execute('stttty -noecho')
sh: stttty: command not found
nil	exit	127
> 
```

There is no separate of `stdout` and `stderr` with PTY based processes.   And I'm hopeful Windows PTYs will work without further configuration.

lua-mode.el
jdtsmith: I hadn't thought of that.  Interestingly, for me I get:

```lua
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
No entry for terminal type "emacs";
using dumb terminal settings.
> io.write('asdf')
asdffile (0x7fff9d8f21a8)
> 
```

But the old version will fail for prompts with more information after them, as in the Shell itself.  We have to tie either to the bol or eol; most comint-major modes with shells choose the former.

jdtsmith:
> Hopefully, one last round of questions. I am sorry it takes forever, but it is just that every rewrite or functionality change has the potential of generating more questions and doubts. Smaller changes are much easier and faster to get in.

No worries; it's a lot of changes and it's best if you understand and are comfortable with them.  Take your time. I'm "officially" done with any major changes to this PR, other than any fixups we decide on, and of course re-sanitizing the history.  Sorry if the ground shifted from under you during this!

jdtsmith:
Any further progress on your review @immerrr?  Let me know if anything needs attention.

immerrr:
Apologies for the delay, last couple of weeks have been crazy. I'll try to have a proper look this weekend.

lua-mode.el
```diff
@@ -1754,6 +1848,151 @@ If `lua-process' is nil or dead, start a new process first."
...
+  (setq lua-shell-redirected-output
+	(with-current-buffer lua-shell-output-buffer
+	  (buffer-substring-no-properties
+	   (point-min) (point-max)))))
+
+(defun lua-mimic-whitespace (string completions)
+  "Reproduce the whitespace pattern of the initial completion string. 
+Lua ignores whitespace, so we must complete across it.  Assumes
+string does not start with whitespace and that all completions start with
+the string."
```
immerrr: I thought that when you have `foo[bar].baz.qux.quu` and you want to complete `quu` your list of completions should have completions starting with `quu` only. At least that's what I'm used to when I use completions in other modes. Is this mimic-whitespace thing related to caching?

scripts/emacs_lua_complete.lua
```diff
@@ -0,0 +1,79 @@
...
+      for k,v in pairs(q.entries) do
+	 if type(k) == "string" and k:sub(1,1) ~= "_" then 
+	    pre = q.pre .. "." .. k
+	    print(pre)
+	    cnt = cnt + 1
+	    if cnt >= limit then break end
+	    if type(v) == "table" then queue:push({pre = pre,entries = v}) end
+	 end
+      end
+   end
```
immerrr: Wait, I just realized, this code traversing the tables beyond the initial completion, e.g. you request
`foo.b` and with `foo.bar` being a valid completion option, it would also go ahead and complete it as `foo.bar.baz`, `foo.bar.qux` and so on, right? If my understanding is correct, at the risk of sounding rude when the implementation has already come this far, I'd say that's too much to my liking. I am accustomed to step-by-step completions, e.g. when you complete one level, input dot and complete the next one. Is there a performance reason to fetch completions in batches of 1000s?

And just out of curiosity, do you have a precedent for this kind of behaviour in other Emacs modes to see if maybe I am lagging behind the Emacs innovation frontier?

scripts/emacs_lua_complete.lua
```diff
@@ -0,0 +1,79 @@
...
+   local g,pre = globals, ""
+   for i = 1,#parts do
+      if not g then break end	-- should always exist
+      if i < #parts then -- keep descending
+	 pre = (i > 1 and pre.."." or "") .. parts[i]
+	 g = g[parts[i]]
+      else  -- final part; seed the queue with any matching tables
+	 for k,v in pairs(g) do
+	    if type(k) == "string" and k:sub(1,1) ~= "_" and k:find('^'..parts[i]) then
+	       local dpath=(i > 1 and pre .. "." or "") .. k
```
immerrr: Does this code handle cases when `k` contains non-alphanumeric characters? E.g. whitespace or dots

scripts/emacs_lua_complete.lua
```diff
@@ -0,0 +1,79 @@
...
+   -- Descend the tree to the limit of dotted parts
+   local g,pre = globals, ""
+   for i = 1,#parts do
+      if not g then break end	-- should always exist
+      if i < #parts then -- keep descending
+	 pre = (i > 1 and pre.."." or "") .. parts[i]
+	 g = g[parts[i]]
+      else  -- final part; seed the queue with any matching tables
+	 for k,v in pairs(g) do
+	    if type(k) == "string" and k:sub(1,1) ~= "_" and k:find('^'..parts[i]) then
```
immerrr: Also, If I'm reading correctly, this ignores any keys that start with underscore to avoid "special" keys. In Python at least it's common to want them shown at least in some situations, so usually it is a configuration parameter on the Emacs side. Do you think it's worth to let underscore-prefixed keys through at least if `parts[i]` starts with an underscore?

lua-mode.el
```diff
@@ -1623,10 +1634,18 @@ This function just searches for a `end' at the beginning of a line."
...
+		     (file-name-directory load-file-name)))
+  "Path to the associate lua completion function definition.")
 (defvar lua-process-init-code
   (mapconcat
    'identity
-   '("local loadstring = loadstring or load"
+   `(
+     ,(concat "dofile(\"" lua-process-complete-code-filename "\")")
+     "os.execute('stty -echo -onlcr')"
+     "local loadstring = loadstring or load"
```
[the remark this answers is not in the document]
jdtsmith: I used to have a finite loop with a timed `accept-process-output`, but removed it.  I don't think that that timeout does quite what we want, in the sense that if output continues to accrue (but no prompt has yet appeared), we want to keep waiting for it.  It also is brittle to select a specific timeout period (2s, 5s, etc.); what if a given command has a long delay of some sort?  What if it's a very low power system (older Raspberry PI, for example)?  In the end I decided to punt on a specific choice and consider that `C-g` is a perfectly fine option for a hung interactive shell.

lua-mode.el
```diff
@@ -1718,17 +1765,64 @@ When called interactively, switch to the process buffer."
...
+    (with-current-buffer lua-process-buffer
+      (unless (buffer-live-p lua-shell-output-buffer)
+	(setq lua-shell-output-buffer
+	      (get-buffer-create
+	       (generate-new-buffer lua-shell-output-buffer-name))))
+      (with-current-buffer lua-shell-output-buffer (erase-buffer))
+      (lua-send-string command lua-shell-output-buffer process)
+      (while (not comint-redirect-completed)
+	(accept-process-output process))
+      (comint-redirect-cleanup))))
```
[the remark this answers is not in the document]
jdtsmith: Agreed, fix pushed.

lua-mode.el
```diff
@@ -1718,17 +1765,64 @@ When called interactively, switch to the process buffer."
...
+  "Send STR to the Lua process, possibly via dofile.
+If necessary, start a new process first.  If optional argument
+REDIRECT-BUFFER is passed, redirect command output to that
+buffer, which calls comint-redirect-hook when the output is
+complete.  If PROCESS not passed, get or create a process."
+  (let (file
+	(command str)
+	(process (or process (lua-get-create-process)))
+	(comint-redirect-perform-sanity-check nil))
+    (when (> (length str) lua-shell-maximum-command-length)
```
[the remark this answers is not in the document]
jdtsmith: That's right.  My testing indicated that there is a maximum command length, and in fact the length of the temp file path on a given emacs install was one explanation for why tests would confusingly fail on Travis while working on our local systems. But I never found this documented.  If indeed there is a maximum length, this could be useful for future interactions with the shell of any kind.

lua-mode.el
```diff
@@ -1718,17 +1765,64 @@ When called interactively, switch to the process buffer."
...
+      (with-temp-file (lua-shell-temp-file)
+	(erase-buffer)
+	(insert command))
+      (setq command (concat "dofile(\"" lua-shell-temp-file "\")")))
+    (setq lua-shell-last-command command)
```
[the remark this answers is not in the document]
jdtsmith: Not anymore; shall I remove it?

lua-mode.el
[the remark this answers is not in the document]
immerrr: What I'm saying that `lua-shell-maximum-command-length` as variable name could be interpreted as the limit to what the user can enter in Lua shell, whereas it is a limit specific to interprocess communication, to which there is already a solution. So now it's really a threshold after which the command is no longer sent via stdin but via temp-file, hence I'm wondering if there's a better name that makes sense to you.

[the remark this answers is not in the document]
jdtsmith: I switched to `lua-shell-maximum-sent-length` and documented as specific to Emacs sending commands to Lua (which I'm not sure if true but better represents our state of knowledge).

lua-mode.el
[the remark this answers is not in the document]
jdtsmith: Removed

lua-mode.el
```diff
@@ -1754,6 +1848,151 @@ If `lua-process' is nil or dead, start a new process first."
           (lua-send-region start end)
         (error "Not on a function definition")))))
 
+(defsubst lua-completion-trim-input (str)
+  (format "'%s'" (string-trim str)))
```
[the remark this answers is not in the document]
jdtsmith: Hmmm.  I tried substituting that and the tests failed.  I think this just needs quote wrapping?

lua-mode.el
```diff
@@ -1754,6 +1848,151 @@ If `lua-process' is nil or dead, start a new process first."
...
+(defvar lua-local-require-completions t
+  "During completion, scan file for local require calls for context.")
+
+(defvar lua-local-require-regexp
+  "^local\\s-+\\([^ \n]+\\)\\s-*=\\s-*require\\s-*(?\\s-*\\(['\"]?\\)\\([^)\n]+\\)\\2"
+  "A regexp to match lines where a library is required and put in a local.")
+
+(defvar lua-top-level-local-regexp
+  "^local\\s-+\\([^ \n]+\\)\\s-*="
+  "A regexp to match top-level local definitions")
```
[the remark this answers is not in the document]
jdtsmith: This was inherited from the initial PR; certainly open to contributions for readability.

lua-mode.el
jdtsmith: This is so you can complete things like:

```lua
foo[bar].
   baz.qux
.q[M-Tab]
```

which is perfectly valid lua.  Since completion looks for exact matches (whitespace and all), we first remove the whitespace, perform the completion, then reproduce the original.

scripts/emacs_lua_complete.lua
jdtsmith: Yes, and this took some real doing, it's the whole point of the breadth-first search.  This is the strongly recommended approach to completion by the completion-at-point facility.  The reason is that there are various completion "styles", and the completion backend is supposed to be agnostic about them, returning a "full list" and letting the completion style sort out what it wants.  For example, [initials](https://www.gnu.org/software/emacs/manual/html_node/emacs/Completion-Styles.html).  On that topic the `completion-at-point-functions` docs include:

> NOTE: These functions should be cheap to run since they’re sometimes
> run from ‘post-command-hook’; and they should ideally only choose
> which kind of completion table to use, and not pre-filter it based
> on the current text between START and END (e.g., they should not
> obey ‘completion-styles’).

and in the manual:

> This specifies an all-completions operation. The function should return a list of all possible completions of the specified string. 

The advantage here is that you can prefetch the deep tree (which is often only dozens deep), and then it is cached for subsequent completes, and all the tests and other calls the completion framework make.  It is sorted so that the "first after dot" is up front, so basically you can ignore it.  It _radically_ cuts down on the back and forth traffic to the lua shell.  

Have you tried it out?  Effectively it does the same thing: complete to the next dot (unless there is only ONE completion beyond that dot, which would be rare).  But far more efficiently and with the option to try some crazy completion styles with company mode, etc.

scripts/emacs_lua_complete.lua
jdtsmith: It would, but in the `parts` that are passed in have already had white space removed from between the dots, for reasons mentioned earlier (lua ignores it!).   And there can't be dots, since that would create a new completion part.

scripts/emacs_lua_complete.lua
jdtsmith: Yes, `_key` is ignored.  My experience is in lua this is a warning for "don't go there, internal variables/methods", but we could have an optional variable to include it (passing in as another parameter to the completion lua function). Two underscores are used for all the metatable stuff, so that probably should be trimmed to avoid circularly connected trees.  This should probably be done, but I'd prefer to put it in another PR.

immerrr: Ok, I see the point in not returning them by default, but would you say that if I enter `mytable.__meta` I should receive `__metatable` as a completion candidate?

scripts/emacs_lua_complete.lua
immerrr: Ah, right, sorry, like I said, too much time away from the code and I start forgetting things

lua-mode.el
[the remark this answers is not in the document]
immerrr: :+1:

lua-mode.el
immerrr: Yes, it is valid lua. What I'm asking is a bit different.

Given `foo.b` you can either complete as "short completions", `["bar", "baz"]`, anchored at the beginning of `b` part, or you can return "long completions", `["foo.bar", "foo.baz"]`, anchored at the beginning of `foo`. Your approach clearly goes for the second way, because the first one avoids the "mimic whitespace" problem altogether, because it only returns the last component which cannot contain whitespace. "short completions" approach seems simpler, and I vaguely remember there being a reason for doing "long completions", just can't remember what was it.

scripts/emacs_lua_complete.lua
```diff
@@ -0,0 +1,79 @@
...
+   end
+
+   -- Mirror the globals table
+   local globals,cnt,limit = {}, 0, limit or 2000
+   for k,v in pairs(_G) do globals[k] = v end
+   
+   -- Add libs and locals
+   if libs then
+      for _,v in ipairs(libs) do
+	 local good,lib=pcall(require,v.lib)
```
immerrr: When trying out stuff locally I have realized that this probably doesn't reload the changes that were made after the first completion attempt. Should we force a reload here, WDYT?

scripts/emacs_lua_complete.lua
jdtsmith: Hmmm... I suppose so for people hacking on modules.  I guess I worry about metatable creating circularly linked trees.  But it cuts off the search so it's not the end of the world.  I could turn it back on by default and see how it works.  In my limited experience (with the `hs` module), it adds a lot of unwanted noise.

scripts/emacs_lua_complete.lua
jdtsmith: It's only if the last completion was a substring of the current completion string (modulo whitespace) that the caching is used. So I don't see how you could get into trouble.  Maybe if you reuse the same local variable but load a different library that might be problematic?  Maybe we should make an explicit binding to drop the last saved completion by setting `lua-last-completed` to nil?

lua-mode.el
jdtsmith: Aha, I see.  Yes, the whitespace fixing is a side effect of supporting long completions.  But you still have to identify the entire expression (whitespace and all).  Short completions were actually (silently) re-querying lua and failing before, since one of the operations of the completions framework is to ask "is this completion string "abc.def" a completion for original expression "abc"?  But the real reason is to make caching and other completions styles possible.  Completions in this context need both a fragment being completed ("abc"), and the list of completions ("abc.def", "abc.xyz").  Making the fragment an empty string means you can't cache it or use other non-trivial styles.

_(below I mistakenly thought you were referring to the depth of the search with short vs. long, but leave it here as supplemental explanation:)_
Short completions are a subset of long completions, and long completions (ideally _every possible completion below this point_, which is what all the completion docs urge you on to) are the preferred way to do completions to work with other styles.  Give the `initials` completion style a try to see what I mean.  It's pretty cool shorthand (especially with company-mode), which simply would not work with "single key at a time" shell querying.  Caching also works best when you pre-cache the `a.b.c.d.e`'s.

scripts/emacs_lua_complete.lua
immerrr: I'm not referring to the `lua-last-completed` caching, I'm talking about the "caching" mechanism behind Lua's `require` function. IIRC it won't reload a library if you `require` it for the second time from the same interpreter.

scripts/emacs_lua_complete.lua
immerrr: We can keep filtering out the special keys in the breadth search, this only applies to when there is a prefix to filter that indicates that the user wants to complete an underscore-prefixed key.

scripts/emacs_lua_complete.lua
jdtsmith: Oh I see.  I thought `require` checked for changes on disk?  Would `dofile` work if not?

scripts/emacs_lua_complete.lua
jdtsmith: I see, so if they explicitly complete on "_xxx[M-Tab]".  That seems reasonable. Just pushed a commit, and took the opportunity to remove the regex based match, since it could be fooled by special characters in the completion fragment.  So it doesn't offer _ keys by default, but if you start with _, it will give you everything.

jdtsmith: Let me know if this works as you'd expect for _XXX keys.

scripts/emacs_lua_complete.lua
immerrr: I think it should be easier to consult `require.loaded` table and remove an entry from there than reimplementing `require`'s path lookups if I'm reading [the doc](http://www.lua.org/manual/5.1/manual.html#pdf-require) correctly.

scripts/emacs_lua_complete.lua
immerrr: Yeah, looks fine.

scripts/emacs_lua_complete.lua
jdtsmith: This seems to me to be beyond the scope of `lua-mode`, since people need to come up with some solution for reloading modules which are changing on disk anyway, even for just _running code which uses the module_.  

I'm suprised `require` doesn't offer changed file reloading.  But since it doesn't, most people would reasonably expect that until they actively "re-require" the module (by clearing it from `package.loaded`, for example), it would remain unchanged.  If `lua-mode` silently re-requires modules in the background through the course of completion, that would invalidate that likely assumption, and result in confusion.  It could also be a big performance hit for large modules to constantly reload them.  

I'm inclined to leave this one to the user to solve (as they already must!).  Or are we looking at it differently?

immerrr: Ok, let's leave this for posterity. You are right that live reloading modules as a side-effect of completions in the subshell should not happen as it is very surprising.

But in general, that's the thing with completion, the user doesn't really care about how the completions are done. I can say for myself, if I change a module in my project I would expect the new changes to be listed in completions right away. I don't care where do they come from, are they cached or are they reloaded on every request (provided it's fast enough). Probably piggybacking the completion functionality on the running Lua inferior shell should be rethought in the future.

immerrr:
Let me have another go over the comments...

emacs_lua_complete.lua
immerrr: No, what I'm saying is that you are basically doing
```
for i = 1, #parts do
    current_part = parts[i]
```
whereas `ipairs` will do that for you automatically
```
for i, current_part in ipairs(parts)
```

emacs_lua_complete.lua
immerrr: Ok. I don't quite like that the functions only output is a side-effect (printing to stdout) and it is impossible to tell from the outside whether or not the function has finished (you can see some lines, but don't know whether or not it is the end), but we can address it later.

emacs_lua_complete.lua
```diff
@@ -0,0 +1,81 @@
...
+	 end
+      end
+   end
+
+   -- Perform a limited breadth-first traversal of the queued tables
+   while queue.head < queue.tail and cnt < limit do
+      local q=queue:pop()
+      for k,v in pairs(q.entries) do
+	 if type(k) == "string" and k:sub(1,1) ~= "_" then 
+	    pre = q.pre.."."..k
```
immerrr: Remember you told me that when keys come from completions, they are verified that they don't contain whitespaces or dots? Here, if I'm not mistaken, `k` comes from Lua tables and no verification is done, and thus `k` can contain whatever and produce completions that are invalid syntax or don't work as expected. WDYT?

emacs_lua_complete.lua
immerrr: Yeah, let's please add whitespace around binary operators and newlines where necessary, now that there's no necessity to keep the size to the minimum.

And things like
```
    local globals,cnt,limit={},0,limit or 2000 
```
could probably use a touch to clarify (`limit` assignment in particular catches me off guard every time I return to this piece of code).

emacs_lua_complete.lua
immerrr: Btw, I mean the entire function that generates completions could be changed to produce an iterator, not just this loop. And then it could be consumed by another outside function that would terminate the iteration when the limit is reached. And then another one that decides how do we communicate that table to the parent process. But like I said, it is just a refactoring, we can address that later.

lua-mode.el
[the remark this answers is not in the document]
immerrr: `lua-make-lua-string` does that: it quote wraps things and escapes few special characters. I'll send in the change.

lua-mode.el
immerrr: Ok, let's see how it works in the wild.

scripts/emacs_lua_complete.lua
immerrr: It is not that I am particularly against "deep" completions, although admittedly I prefer the shorter version and no craziness. 

It's just that we already do completions on a live interpreter that could produce some side-effects, we have to be careful about it. With "deep" completions it takes one object hidden somewhere within the initial 1000 element set to have a `__pairs` method doing something crazy to have a really bad experience with this. And I'm don't mean malevolent, just crazy. With "short" completions I as a user have some control over whose keys I want to be iterated in the background and whose not. With "deep" completions I am at the mercy of the code of the library, or I have to verify each library with my own eyes.

> It radically cuts down on the back and forth traffic to the lua shell.

Many Pandora boxes have been opened with good intentions to optimize things. Spectre/Meltdown family of CPU vulnerabilities immediately come to mind, but surely there are more.

immerrr:
I have added [pr-148-suggestions branch](https://github.com/immerrr/lua-mode/commits/pr-148-suggestions), feel free to have a look.

scripts/emacs_lua_complete.lua
jdtsmith: OK, thanks.  One future solution is we could provide a convenient shortcut for "dump the `package.loaded` of all the modules `require`d in this file".  That would then be of general use for better completion and inheriting new Lua behavior on running your code. And yes, parsing the Lua code directly is more flexible (and way more complicated, I know from experience with another major mode of mine for the IDL language).

emacs_lua_complete.lua
jdtsmith: Good point.  Yeah any valid key in Lua can be in `k`.  But you can't reference such keys with dot-syntax (I just tried).  Since we are only parsing and completing dot-syntax I think we can get away with it.  It would be better of course to parse `a[" foo "].b[" c"]` style things too, but that's a new can of worms.  If that ever comes to pass, you'd trim and recreate whitespace around dots, but NOT within brackets (since there it is significant and must be preserved).  I guess it might not be that hard to take an arbitrary expression with brackets and dots (and comments) and turn it into a series of parts with all the ignorable whitespace removed and all the relevant whitespace retained.  Right now it's this simple bit:

```lua
 (mapcar 'lua-completion-trim-input (split-string expr "\\."))
```
Wouldn't be terrible to update this to take the expression and remove comments, find brackets etc.  Right now `lua-start-of-expression` doesn't handle brackets gracefully.

emacs_lua_complete.lua
jdtsmith: Good point, fixed.

emacs_lua_complete.lua
jdtsmith: Interesting idea.  That's a bit beyond my present Lua familiarity (just started with it a few weeks ago).  Hmmm... do you imagine having a special code to signal end of output?  Unless you open pipes etc. there is only one way to communicate with Emacs from asynchronous processes: printing to stdout.  The former version of writing to file and waiting an arbitrary time delay then reading it in was very brittle.  But I suppose you could write to temp file, then write the name of that file on stdout and let Emacs read the file name and load and parse the file.   Maybe a decent blended approach.

scripts/emacs_lua_complete.lua
jdtsmith: Hmm, I hadn't thought of rogue objects overriding __pairs.  Can you mock up an example?  Maybe there's a way to protect against that.  Deep completions are now a fundamental part of modern Emacs completion framework, for better or worse.  Given a starting fragment, you are ideally to return all possible completions of that fragment without prejudice, which enables many modern forms of completion that many people use.  If this is a deal-breaker for you, I really do wish you had voiced the concerns back when I first proposed a depth-first search of the globals tree, and before I spent ~days implementing it.

emacs_lua_complete.lua
immerrr: I think the simplest solution is to just filter out any keys that are not alphanumeric-plus-underscore here.

lua-mode.el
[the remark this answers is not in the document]
jdtsmith: Be sure to `make test`; on a quick substitution I found it failed.

scripts/emacs_lua_complete.lua
immerrr: If you don't mind, I'll respond to this as a generic comment, as this is a quite deep topic (no pun intended) and it becomes increasingly difficult to keep track of all the "line comment threads" as lines are updated and the comments themselves become outdated.

immerrr:
> Hmm, I hadn't thought of rogue objects overriding __pairs. Can you mock up an example? Maybe there's a way to protect against that.

I have tried searching online, the things I was able to find were not exactly rogue, but you can probably get the idea:
- few examples of filesystem-like abstractions, like [this one](http://lua-users.org/lists/lua-l/2006-04/msg00518.html)
- [one](https://github.com/OgreDee/Dee_LuaADT/blob/ad52e53c59c4a33239679e19405837d730396792/tuple.lua#L33) that actually returns an error "protecting" the stored keys 
- [one place where the table is primed from cache when accessing keys](https://github.com/abgood/crazy/blob/3e56535f14289f854623fce4ccbc117a5912a263/tools/w3x2lni/script/map-builder/archive.lua#L136)

There's a [function](https://github.com/pkulchenko/serpent/blob/master/src/serpent.lua#L6) that I've seen on various occasions that is supposed to help avoid the `__pairs` metamethod.

```lua
M = {
   __pairs = function (tbl)
      return pairs({1, 2, 3})
   end
}

foo = setmetatable({foo=1, bar=2}, M)

[[
> for k, v in pairs(foo) do print(k, v) end
1	1
2	2
3	3
]]

local mypairs = function(t) return next, t end

[[
> for k, v in mypairs(foo) do print(k, v) end
bar	2
foo	1
]]
```

> Deep completions are now a fundamental part of modern Emacs completion framework, for better or worse.

I'd like to apologize again, I should have brought up the concern that is about to follow when you first brought it up [here](https://github.com/immerrr/lua-mode/pull/148#discussion_r292078410). I don't think I was able to follow your thought as it went from the documentation to the idea that deep completions are the way to go.

You mentioned the phrase "full list" as a hint that sub-expansions (aka "deep" completions) are expected, but I don't quite see how "full list" means "descend eagerly and recursively". It sounds to me that there is a restriction that what Emacs works with should be a closed set (so, "set of natural numbers" doesn't work because it is unbounded) and then sometimes you should be able to return it in its entirety.

You have to keep in mind that Emacs uses the same completion framework for intelligent code-based completions and for example filename completions. In filename context it makes sense that they explicitly mention "a large list of possible completion alternatives", because you can have hundreds and thousands of files in one directory and completion is meant to help you find the one you need. And speaking of file completions, from my experience the general approach is not go inside directories automatically _unless_ there is a specific and _closed_ set of full paths to work with, e.g. all files in a git repo or in an archive. 

You were mentioning `initials` completion style, that I think suits "shallow" one-level completions too, e.g. when you have a class with a hundred-ish methods looking like `do_this_fancy_action` and with `initials` enabled you can actually complete it as `dtfa`. It is easy to see how the same thing would apply to filename expansion.

In fact I can't remember having seen a completion framework that would do "deep" and "unbounded" completions. That includes various Emacs major modes, Bash completions for various commands that had custom completions implemented, and a short period of my career when I did C++ in MS Visual Studio. Do you have an example where a completion framework would eagerly descend into the completed objects?

> Given a starting fragment, you are ideally to return all possible completions of that fragment without prejudice, which enables many modern forms of completion that many people use.

I completely agree with this part, but again it all ultimately depends on what do you mean by "possible completions": whether it is "keys of a single table" or "keys of a single table and keys of any nested tables".

> If this is a deal-breaker for you, I really do wish you had voiced the concerns back when I first proposed a depth-first search of the globals tree, and before I spent ~days implementing it.

I am incredibly sorry about your experience, whether we find a way for this to be included like this or not. I commend your enthusiasm in tackling such a difficult issue and your devotion in seeing it through. Without a doubt, open-source community needs more people like you :) 

For me it is probably the first time that I experience the scope of a contribution changing so drastically. Admittedly, my time was and is still scarce, so I could seem on and off, but in those rare moments when I was trying to wrap my head about the implications of the changes, sometimes it felt as if there was always another piece of functionality waiting around the corner.

I presume the breadth-first implementation came somewhere between [this](https://github.com/immerrr/lua-mode/pull/147#issuecomment-493235232) and [this](https://github.com/immerrr/lua-mode/pull/147#issuecomment-495461057) comments. To give you some context, [in this message](https://github.com/immerrr/lua-mode/pull/147#issuecomment-493771913) it sounded to me like you agreed to postpone drastic changes until after the initial implementation is merged, so I didn't expect the changes in behaviour. Again, regardless of that, I should have been clearer in communication and more attentive to the changes themselves, I'm sorry about that.

immerrr:
With that being said, I'd propose the following plan:

- I'll send a message to `emacs-devel` mailing list to ask for clarification about "deep" completion trees and how to support them better
- I'd probably suggest to separate completion into "tiers", something like 
  -  `:globals-only` (the safest, no `require` is happening in the background so no side effects are expected, ideally that would also be the default)
  - `:globals-and-libs` (there could be side effects, but you are more in control)
  - `:globals-and-libs-nested` (the current implementation and the most feature-rich one)
- In addition to that, there is an option to use the "custom `pairs` function" approach can counteract the risk of running into "crazy metatables" at the expense of not having the best completions for objects using those crazy metatables

Please, let me know what do you think before doing the changes.

jdtsmith:
> There's a [function](https://github.com/pkulchenko/serpent/blob/master/src/serpent.lua#L6) that I've seen on various occasions that is supposed to help avoid the `__pairs` metamethod.
> 
> ```lua
> M = {
>    __pairs = function (tbl)
>       return pairs({1, 2, 3})
>    end
> }
> 
> foo = setmetatable({foo=1, bar=2}, M)
> 
> [[
> > for k, v in pairs(foo) do print(k, v) end
> 1	1
> 2	2
> 3	3
> ]]
> 
> local mypairs = function(t) return next, t end
> 
> [[
> > for k, v in mypairs(foo) do print(k, v) end
> bar	2
> foo	1
> ]]
> ```

Does a virtual rewrite of keys like this actually break completion?  I.e. aren't virtual keys "as good as" real keys, for all intents and purposes?

> > Deep completions are now a fundamental part of modern Emacs completion framework, for better or worse.
> 
> I'd like to apologize again, I should have brought up the concern that is about to follow when you first brought it up [here](https://github.com/immerrr/lua-mode/pull/148#discussion_r292078410). I don't think I was able to follow your thought as it went from the documentation to the idea that deep completions are the way to go.

Yeah I know, and I do apologize as well.  It came pretty fast as I had some free time and interest in getting a system that works.  I can see how it might have seemed out of the blue.

> You mentioned the phrase "full list" as a hint that sub-expansions (aka "deep" completions) are expected, but I don't quite see how "full list" means "descend eagerly and recursively". It sounds to me that there is a restriction that what Emacs works with should be a closed set (so, "set of natural numbers" doesn't work because it is unbounded) and then sometimes you should be able to return it in its entirety.

> You have to keep in mind that Emacs uses the same completion framework for intelligent code-based completions and for example filename completions. In filename context it makes sense that they explicitly mention "a large list of possible completion alternatives", because you can have hundreds and thousands of files in one directory and completion is meant to help you find the one you need. And speaking of file completions, from my experience the general approach is not go inside directories automatically _unless_ there is a specific and _closed_ set of full paths to work with, e.g. all files in a git repo or in an archive.

Right, so it depends on whether completing `a` to `a.b a.b.c a.b.d a.b.c.e a.b.d.f` is seen as a single entity (like the string "abc"), or a nested structure with natural non-crossable boundaries.  Clearly we have different points of view here.  The set is closed since it is limited to a finite (and configurable) number of entities.  

But even the directory as non-crossable boundary breaks down.  There is precedent even for your example of nested directories.  E.g. in `icomplete`, `C-x C-f /u/l/s/e` returns the completion of `/usr/local/src/emacs`.  And see below for another example that you can try _right now_, without any additional configuration or package.

> You were mentioning `initials` completion style, that I think suits "shallow" one-level completions too, e.g. when you have a class with a hundred-ish methods looking like `do_this_fancy_action` and with `initials` enabled you can actually complete it as `dtfa`. It is easy to see how the same thing would apply to filename expansion.

Fuzzy completion is popular in helm, company-mode, Ido, etc. Quoting a [stack exchange article](https://emacs.stackexchange.com/questions/2671/how-can-i-get-fuzzy-code-completion):

> E.g. if you open an Elisp buffer using Emacs-24.4 and the latest company-mode, if you type (wi-sy company will pop up a completion menu showing `window-system`, `window-system-for-display` and `with-syntax-table`.

The distinction between `-` and `.` (or `/`) would not, IMO, be significant to a general user. 

To make it concrete, here's an example that I am **already** using happily that would not work with shallow "one-level-at-a-time" `initials` style completion (again using the deep Hammerspoon `hs` structure): 

```
hs.ev.ev.ra[M-Tab]
``` 

turns into:

```
hs.eventtap.event.rawFlagMasks
```

Nice, right?  If `hs.ev.ev.ra` is not an "initialism", as described in the `completion-styles` docs, I don't know what is!

Here's the crux: earlier you made an argument that optimizing can lead to suboptimal outcomes; is this a case where you feel that too long of a completion list is somehow "too heavy", so that you are performing an implicit (counter-?) optimization of your own?  

**IMPORTANT POINT HERE:**

I'm trying to understand if your main objection is to 

1. The _hidden completion traffic_ of (up to, though usually far less than) 2000 (or whatever) nested keys, or 
2. The resulting _completion interface_ in which a long list is presented (with the shallowest presented first)

If it's the latter, it's possible we could keep the cached completion breadth-first search backend, but optionally prune it "on demand" to keep the completion interface simpler (but still enabling fuzzy completion of the sort I gave as an example above).  For example, if we presented only one level of completion at a time, would you _care_ that in the background `lua-mode` had done a deep recursive (but length limited) completion search, and cached the results?  As you said before, the user simply doesn't care how and where the completion data came from.  They just want their completions, fast and complete. 

> In fact I can't remember having seen a completion framework that would do "deep" and "unbounded" completions. That includes various Emacs major modes, Bash completions for various commands that had custom completions implemented, and a short period of my career when I did C++ in MS Visual Studio. Do you have an example where a completion framework would eagerly descend into the completed objects?

All the fuzzy matching completion frameworks (many of which are in the mini-buffer) do this.  Ido, icomplete, helm, etc. Granted sometimes they have interfaces to step through "one directory at a time". But behind the scenes, they explore and cache a large part of the search space.  

In fact, _**even  for filepaths**_, you can demonstrate this to yourself right now (courtesy `comint` mode). Try this in a `*lua*` buffer:

```
> file="/us/loc/sh/em[M-Tab]
```

I get:

```
> file="/usr/local/share/emacs/
```

How?  By recursively searching the directory tree for things that match!

> > Given a starting fragment, you are ideally to return all possible completions of that fragment without prejudice, which enables many modern forms of completion that many people use.
> 
> I completely agree with this part, but again it all ultimately depends on what do you mean by "possible completions": whether it is "keys of a single table" or "keys of a single table and keys of any nested tables".

Indeed, that's the crux of the issue.   

> > If this is a deal-breaker for you, I really do wish you had voiced the concerns back when I first proposed a depth-first search of the globals tree, and before I spent ~days implementing it.
> 
> I am incredibly sorry about your experience, whether we find a way for this to be included like this or not. I commend your enthusiasm in tackling such a difficult issue and your devotion in seeing it through. Without a doubt, open-source community needs more people like you :)

It's no worries, it's not my first rodeo :).  I have long maintained (though now on minimal maintenance) a large Emacs major mode of over 35k lines with >25years of history (which has it's own terribly powerful but terribly hackish and non-standard completion framework; hence my interest in modern Emacs-supported completion!).  I know how spending large amounts of time and energy on something can make you feel about it: almost like family. 

> For me it is probably the first time that I experience the scope of a contribution changing so drastically. Admittedly, my time was and is still scarce, so I could seem on and off, but in those rare moments when I was trying to wrap my head about the implications of the changes, sometimes it felt as if there was always another piece of functionality waiting around the corner.

It's a strange feeling I know.  I've certainly had that experience a well: a package I have known and maintained for years is suddenly attracting interest of other developers, even complete forks and ports to a new language.  As anyone I think would, I felt a natural protective resistance.  Eventually I sat with it, and the two things I had to ask myself were: 1) are these features and capabilities ones that I would want for the users? 2) am I willing and able to implement them myself?  If I answered Yes and No, I had to consider whether my own instinctive protectiveness was serving the end goal.  

> I presume the breadth-first implementation came somewhere between [this](https://github.com/immerrr/lua-mode/pull/147#issuecomment-493235232) and [this](https://github.com/immerrr/lua-mode/pull/147#issuecomment-495461057) comments. To give you some context, [in this message](https://github.com/immerrr/lua-mode/pull/147#issuecomment-493771913) it sounded to me like you agreed to postpone drastic changes until after the initial implementation is merged, so I didn't expect the changes in behaviour. Again, regardless of that, I should have been clearer in communication and more attentive to the changes themselves, I'm sorry about that.

I admit I did move ahead without confirmation after waiting with some frustration for many days for you to digest and opine on the suggested plan, and I could have been clearer about my direction.  I had intended to work on this for a ~week, and had many other pressing things I had to move on to.  I could have been clearer about that. And partly it was an exercise to learn more Lua and more about modern Emacs completion (which succeeded, on both counts). 

So if you want to scrap it all and wait for a 3rd person to come along and implement completion in `lua-mode` in another few years, I'll respect that decision.  I don't have enough interest and future potential with Lua to maintain my own fork.  But if that's the case, I think it's worth asking yourself why not one but two different significant efforts to contribute completion, which you have acknowledged before you yourself were unwilling to implement but very interested in having, have withered on the vine.  I really don't mean to offend by this, only to offer some counsel from my own long experience.  We all have our blind spots. 

No rush here to think about all this.  I'm going to put this down entirely until I hear from you again.  I can contribute some modest effort to improving the completion interface if that's your main objection, but will not have the time to completely re-implement what in my view is a well-working and already very function tool!

immerrr:
If you don't mind I'll reorder the quotes so that my message makes a bit more sense when read sequentially.

> It's no worries, it's not my first rodeo :). I have long maintained (though now on minimal maintenance) a large Emacs major mode of over 35k lines with >25years of history (which has its own terribly powerful but terribly hackish and non-standard completion framework; hence my interest in modern Emacs-supported completion!). I know how spending large amounts of time and energy on something can make you feel about it: almost like family.

If anything, that makes your enthusiasm even more admirable. And does not invalidate my point that more people with the attitude like yours would be welcome anywhere in the OSS community :)

> It's a strange feeling I know. I've certainly had that experience a well: a package I have known and maintained for years is suddenly attracting interest of other developers, even complete forks and ports to a new language. As anyone I think would, I felt a natural protective resistance. Eventually I sat with it, and the two things I had to ask myself were: 1) are these features and capabilities ones that I would want for the users? 2) am I willing and able to implement them myself? If I answered Yes and No, I had to consider whether my own instinctive protectiveness was serving the end goal.

Ok, yes, I am a bit protective, but not so much about the code but rather about the user experience. If there was one piece of feedback I kept hearing when asking about what next big thing would help lua-mode people would answer something along the lines of  "it is not broken now, it does what it is supposed to, and that's probably enough". So it kind of reinforced this approach of mine that the "continuity" comes first.

To give an example of continuity, I have the case of the unfortunate lua-mode's default indentation offset of 3 spaces. The majority of Lua code styles out there use either 2 or 4, 3 at least right now seems like a weird historical artifact that got fossilized. With respect to it, the userbase is distributed across 3 major camps:
1. existing users who are happy with the offset of 3
2. existing users who were unhappy with the offset of 3 and customized it
3. new users who have the offset documented as a customization variable and can join either the first or the second camp

Now, I'm reluctant to change the default, because after doing so:
- group 2 won't even notice when they upgrade, they have the default already overridden
- group 3 will have a minuscule improvement: they are installing a new package, so they are reading the docs and customizing it anyway, one less "set-it-and-forget-it" step does not make much of a difference
- group 1 however upon upgrading will see the difference which they don't like (they wanted 3 spaces after all, right?), so now they have to go and _reconfigure_ an already configured package. I cannot tell for others, but I can get pretty annoyed when people break things while trying to fix something that IMO was not broken in the first place. 
 
FWIW if someone came in and rewrote the entire package in a maintainable fashion while keeping the UX and provide some assurance that the users won't experience problems when upgrading I'd be ecstatic to throw out all of my code. If along comes someone who shares my POV about the continuity in the UX and is happy to share the burden of maintainership or even take over, even better. 

> **IMPORTANT POINT HERE:**
> 
> I'm trying to understand if your main objection is to
> 
>     1. The _hidden completion traffic_ of (up to, though usually far less than) 2000 (or whatever) nested keys, or
> 
>     2. The resulting _completion interface_ in which a long list is presented (with the shallowest presented first)
> 
> 
> If it's the latter, it's possible we could keep the cached completion breadth-first search backend, but optionally prune it "on demand" to keep the completion interface simpler (but still enabling fuzzy completion of the sort I gave as an example above). For example, if we presented only one level of completion at a time, would you _care_ that in the background `lua-mode` had done a deep recursive (but length limited) completion search, and cached the results? As you said before, the user simply doesn't care how and where the completion data came from. They just want their completions, fast and complete

With completions the continuity I'm trying to keep is more about "My workstation is safe w.r.t. side-effects that could exist in Lua code, the code is not executed unless I tell it to". And with that I try to avoid the frustration of an old-time user who decides to upgrade and then gets bitten by an absolutely random side effect. If that happened to me, I would most likely feel like the package and indirectly the maintainers would have betrayed my trust. If I were especially annoyed I would make a case of it on the Internet about how it is a potential vulnerability created out of thin air when there was no real necessity.

Naturally, the "no side-effect" invariant is impossible to maintain with completion that runs inside the inferior shell and loads libraries to be able to complete them, but I'm trying to reduce the "vector of attack", so to say, and expanding the completion tree recursively and without restrictions in my eyes does the opposite. And I think I now see how my concern can come off as protectiveness against change in general, sorry about that.

> To make it concrete, here's an example that I am **already** using happily that would not work with shallow "one-level-at-a-time" `initials` style completion (again using the deep Hammerspoon `hs` structure):
> 
> ```
> hs.ev.ev.ra[M-Tab]
> ```
> 
> turns into:
> 
> ```
> hs.eventtap.event.rawFlagMasks
> ```
> 
> Nice, right? If `hs.ev.ev.ra` is not an "initialism", as described in the `completion-styles` docs, I don't know what is!

I think you are on to something with this example and the other ones about filenames and directories. I can confirm that  `file="/us/loc/sh/em` expands to `file="/usr/local/share/emacs/`, and with this you have me convinced that there is a place for "deeper" expansions. One disagreement down, couple more to go :)

But now see how for example for `file="/us/local/s[M-tab]` the search tree is pruned to only 3 alternatives, `/usr/local/share, /usr/local/sbin, /usr/local/src`. So it looks like as if the set of candidates is restricted to all nodes at level 3 of the tree. The code in this PR (please correct me if I'm wrong) in a similar situation would continue recursive expansion and show all the keys below those 3 and all the keys below those child keys and so on. That is the behaviour that doesn't quite sound right.

> Here's the crux: earlier you made an argument that optimizing can lead to suboptimal outcomes; is this a case where you feel that too long of a completion list is somehow "too heavy", so that you are performing an implicit (counter-?) optimization of your own?

Judging from the above, I don't think I have much objection against fuzzy completion or a depth-N (with any specific N > 1) search. You have helped me understand that probably my main concern is when the depth is unbounded. And that brings us to the question to the question of separators and which separators add depth, and which do not.

> The distinction between `-` and `.` (or `/`) would not, IMO, be significant to a general user.

With this, I think I still disagree. Dashes — that come from Elisp I presume — are different from dots, because there is no namespacing in Elisp. All functions are on the same level and it makes sense to return the entire list which is, again, closed. In Lua, the closest analog to the Elisp situation with dashes and the lack of namespaces would be if Lua had all the possible completion alternatives as one big  global table with variables whose names consisted of words separated by underscores. The fundamental difference between an underscore and a dot w.r.t depth is as follows:
- for variables `foo_bar_baz` and `fox_qux_quux`, and `fo` as the string that you want to complete, `foo` or `fox` by themselves are not valid completion results. So it makes sense to show variable names as if they were "fully expanded". Hence there is no notion of "depth", just different number of words in variable names which can be fuzzy completed all the way.
- for a similar table situation with two fully expanded completions being `foo.bar.baz` and `fox.qux.quux`, `foo` and `fox` *can* be a valid completion, so subsequent parts increase depth. IMO filename completion, as implemented right now in comint mode, does the right thing by not providing any alternatives beyond the `foo`/`fox` choice, unless you specifically ask for completing across more than 1 level of depth with smth like `fo.b.b`.

If we agree to do the same in Lua completions, we can reduce the depth of the object tree's traversal and with it the "attack vector" for the unexpected side effects.

> Does a virtual rewrite of keys like this actually break completion? I.e. aren't virtual keys "as good as" real keys, for all intents and purposes?

Virtual rewrite of keys does not break completion, but it opens a wide field for side effects to creep in. Implementing an alternative side-effect-free `pairs` function is one way around this, at the cost of making the completion framework oblivious towards all that metatable magic. 

> I admit I did move ahead without confirmation after waiting with some frustration for many days for you to digest and opine on the suggested plan, and I could have been clearer about my direction. I had intended to work on this for a ~week, and had many other pressing things I had to move on to. I could have been clearer about that. And partly it was an exercise to learn more Lua and more about modern Emacs completion (which succeeded, on both counts).

Again, sorry about the frustration. I'm doing it on a best-effort principle, life and day job come first.
 
> So if you want to scrap it all and wait for a 3rd person to come along and implement completion in `lua-mode` in another few years, I'll respect that decision. I don't have enough interest and future potential with Lua to maintain my own fork. But if that's the case, I think it's worth asking yourself why not one but two different significant efforts to contribute completion, which you have [acknowledged before](https://github.com/immerrr/lua-mode/issues/87) you yourself were unwilling to implement but very interested in having, have withered on the vine. I really don't mean to offend by this, only to offer some counsel from my own long experience. We all have our blind spots.
>
> No rush here to think about all this. I'm going to put this down entirely until I hear from you again. I can contribute some modest effort to improving the completion interface if that's your main objection, but will not have the time to completely re-implement what in my view is a well-working and already very functional tool!

There are 2 blockers that I see right now:
- have melpa recipe updated, so that people that install the newer version don't end up without `emacs_lua_complete.lua` file
- have it disabled by default so that existing users that are used to "no-side-effects" approach are not surprised by it OR (ideally, AND) when completion is enabled, reduce the amount of expansions performed as much as possible (performance considerations come later)

Worst case scenario, if I bore you to the point where you would feel inclined to close this issue and move on with your life, I think with your blessing I'll be able to try to fix those blockers on my own. Hopefully, that would happen sooner than the 3rd person would arrive to the scene.

jdtsmith:
Sorry I never followed up on this.  I do like the idea of "do depth completion only if requested", ala a deep initials-style.  But of course for simplicity, you could still generate and cache the full list from the lua process, but choose only to show limit portions of it.  

In any case, you have my full blessing to hack on this branch and fix the blockers you mention.  Still using this branch with success, but not working with lua too much lately.

PR #151, opened 2019-10-02
Enhance indentation of blocks 

edam:
Added `lua-indent-nested-block-content-align` var, which disables the aligning of content of nested blocks to the opening brace and indents normally.  It defaults to `t`, which is the previous behaviour.

E.g. 1:

With `lua-indent-nested-block-content-align` set `t` (default), then:
```lua
call_some_fn( something, {
                val = 5,
                another = 6,
} )
```

Wth `lua-indent-nested-block-content-align` set `nil`, then:
```lua
call_some_fn( something, {
    val = 5,
    another = 6,
} )
```

E.g. 2:

With `lua-indent-nested-block-content-align` set `t` (default), then:
```lua
local def = {
  some_very_long_name = { fn =
                            function()
                              return true
                            end
  }
}
```

Wth `lua-indent-nested-block-content-align` set `nil`, then:
```lua
local def = {
  some_very_long_name = { fn =
      function()
        return true
      end
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ modules to your Emacs and keeping them up-to-date. Once you have **el-get** set 
 and updating is no more than
 
     <M-x> el-get-update "lua-mode"`
-    
+
 Please, consult with [el-get documentation](https://github.com/dimitri/el-get/blob/master/README.md) for further information.
 
 ### MANUAL INSTALLATION
@@ -53,6 +53,7 @@ The following variables are available for customization (see more via `M-x custo
 
 - Var `lua-indent-level` (default `3`): indentation offset in spaces
 - Var `lua-indent-string-contents` (default `nil`): set to `t` if you like to have contents of multiline strings to be indented like comments
+- Var `lua-indent-nested-block-content-align` (default `t`) set to `nil` to stop aligning the content of nested blocks with the open parenthesis
 - Var `lua-mode-hook`: list of functions to execute when lua-mode is initialized
 - Var `lua-documentation-url` (default `"http://www.lua.org/manual/5.1/manual.html#pdf-"`): base URL for documentation lookup
 - Var `lua-documentation-function` (default `browse-url`): function used to show documentation (`eww` is a viable alternative for Emacs 25)
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ The following variables are available for customization (see more via `M-x custo
 - Var `lua-indent-level` (default `3`): indentation offset in spaces
 - Var `lua-indent-string-contents` (default `nil`): set to `t` if you like to have contents of multiline strings to be indented like comments
 - Var `lua-indent-nested-block-content-align` (default `t`) set to `nil` to stop aligning the content of nested blocks with the open parenthesis
+- Var `lua-indent-close-paren-align` (defaut `t`) set to `t` to align close parenthesis with the open parenthesis rather than with the beginning of the line
 - Var `lua-mode-hook`: list of functions to execute when lua-mode is initialized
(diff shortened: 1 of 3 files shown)


immerrr:
Hey, thank you for the contribution! 

It has docs, maintains backward compability, and the change is very focused and self-contained. Great stuff! Could you please add some tests to ensure this behaviour is not lost in any potential refactoring down the road?

edam:
I've also added `lua-indent-close-paren-align` var, to change the way the close parenthesis are aligned, so that they always align with the beginning of the line with the open parenthesis on.  It defaults to `t`, which is the previous behaviour.

E.g.,

With `lua-indent-close-paren-align` set to `t` (the default), then:
```lua
local foo = setmetatable( {
    a = 4,
    b = 5,
                          }, {
    __index = some_func,
} )
```
With `lua-indent-close-paren-align` set to `nil`, then:
```lua
local foo = setmetatable( {
    a = 4,
    b = 5,
}, {
    __index = some_func,
} )
```

immerrr:
Hey, awesome stuff with the tests, but I'm not sure if we need the `close-paren` variable: if you are not aligning the block, there is not much sense in aligning the closing brace.

edam:
The `nested-block-content-align` variable only affects the content of nested blocks.  The `lua-indent-close-paren-align` variable does something different. 

E.g.,

Default settings:
```lua
call_some_fn( something, {
                val = 5,
                val = 7,
                         }, {
                val = 6,
} )
```

With `lua-indent-nested-block-content-align` set to `nil`:
```lua
call_some_fn( something, {
    val = 5,
    val = 7,
                         }, {
    val = 6,
} )
```

With `lua-indent-close-paren-align` set to `nil`:
```lua
call_some_fn( something, {
                val = 5,
                val = 7,
}, {
                val = 6,
} )
```

With *both* `lua-indent-nested-block-content-align` *and* `lua-indent-close-paren-align` set to `nil`:
```lua
call_some_fn( something, {
    val = 5,
    val = 7,
}, {
    val = 6,
} )
```

edam:
Would you like me to combine `lua-indent-nested-block-content-align` and `lua-indent-close-paren-align` in to a single variable?

immerrr:
> Would you like me to combine lua-indent-nested-block-content-align and lua-indent-close-paren-align in to a single variable?

Hmm... I think I missed one case here. 

My original idea stemmed from the fact that there could be different preferences w.r.t. where should the table content go:
- either aligned to the opening brace plus indented (current behaviour in lua-mode)
- or indented as the line of the opening brace plus one offset (the most popular suggested alternative)

But I couldn't find a reason for the closing brace to be anywhere else but at content indentation minus 1 indentation offset. This was about the case when there is no potential alignment inside the table, i.e. when the inside table elements start on the subsequent line,
```
foo(bar, {
    1,
    2,
})
```

But then your comment made me realize that there are still cases when the first element in the nested table is on the same row as the opening brace and it creates an alignment dilemma, so to speak, even without the outer function call:
```lua
-- this looks fine
foo = {1,
       2,
       3}

-- but the following two look strange
foo = {1,
       2,
       3
}

foo = {1,
       2,
       3
      }
```

So I guess it would make sense to introduce a variable for people to "pick their poison" in cases when table elements are aligned, but the close brace is on a separate line.

edam:
So are you saying to leave it as it is, so the user has the choice? 🤔

edam:
So are you saying to leave it as it is, so the user has the choice? 🤔

immerrr:
Sorry for the delay here. 

I'm going to merge it. Thank you for your work and for your patience!

immerrr:
Hey!

Looks like `lua-indent-nested-block-content-align` setting applies not only to nested blocks, but to any alignment in general. 
```lua
foobar(123,
   456)
```
In this example there are no nested blocks, but since alignment is disabled, 456 is indented instead.

PR #156, opened 2019-11-04
Add lua-send-string-max-size and lua-send-region-chunked

immerrr:
This PR adds `lua-send-string-max-size` variable that limits the size of the data that's sent to subprocess in one go. It also introduces `lua-send-region-chunked` function that sends the code in chunks that don't exceed given size to sidestep the size limitation.


astoff:
FWIW, the variable `comint-max-line-length`, new in Emacs 28, seems related to this.

PR #160, opened 2019-11-10
Fix typo

tarsius:
Unfortunately I missed that before.  Because `codespell` did not auto-correct because it wanted me to chose from two possible corrections.


immerrr:
Thanks!

PR #161, opened 2019-11-28
Use proper lexical-bindings instead of lexical-let*

tarsius:
... and other small improvements.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -86,7 +86,7 @@
 
 ;;; Code:
 (eval-when-compile
-  (require 'cl))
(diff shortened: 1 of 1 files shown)


immerrr:
Hi, thank you for the contribution, looks great! Some CI is failing there for older Emacsen, some functions were probably only made available recently.

tarsius:
Sorry I missed that. Fixed by using `nth` instead of `caddr` and `cadddr`.

immerrr:
Awesome stuff, thanks!

PR #165, opened 2020-04-04
Don't use non-deterministic %d with git's export-subst

infinisil:
This prevents GitHub's tarballs from changing depending on whether a
branch points to the commit or not. E.g. on latest master, %d would
evaluate to "master", whereas once master is updated again, the same
commit wouldn't be master anymore, meaning %d evaluates to "".

This is the cause for PRs https://github.com/NixOS/nixpkgs/pull/56123 and
https://github.com/NixOS/nixpkgs/pull/84277

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -36,7 +36,7 @@
 
 ;; This field is expanded to commit SHA, date & associated heads/tags during
 ;; archive creation.
-;; Revision: $Format:%h (%cD %d)$
+;; Revision: $Format:%h (%cD)$
 ;;
 
 ;;; Commentary:


veprbl:
Seems like we independently came to the same solution #166

lua-mode.el
```diff
@@ -36,7 +36,7 @@
 
 ;; This field is expanded to commit SHA, date & associated heads/tags during
 ;; archive creation.
```
immerrr, line 38: I guess we need to update the comment here as well then, right?
```suggestion
;; This field is expanded to commit SHA & date during archive creation.
```

veprbl: This is already done in my pr

immerrr: Thank you for the info!

immerrr:
Whereas this PR came earlier, I have merged #166 that does the same thing but is just slightly more complete. I didn't want to get distracted and forget about this issue, which happened more times than I'd be proud of.

I hope that's OK with you, but let me know if you'd like me to clarify the situation in the sources.

PR #166, opened 2020-04-05
Fix reproducibility issue caused by %d pattern in export-subst

veprbl:
The current revision string format results in release tarballs that have different content depending on what branches are pointing to a given revision. This is an issue for distributions that download such tarballs (e.g. from GitHub) and validate them agains reference hashes.

See downstream issues NixOS/nixpkgs#71948 and NixOS/nixpkgs#84312

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -34,9 +34,9 @@
 
 ;; Keywords: languages, processes, tools
 
-;; This field is expanded to commit SHA, date & associated heads/tags during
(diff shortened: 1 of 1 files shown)


immerrr:
Merged, thank you!

And apologies for the inconveniences with the downstream packaging.

Infinisil:
Tbh, this is rather a problem with git, it should never have allowed %d for this purpose, I think this is the only formatter that's not deterministic from the commit.

PR #170, opened 2020-05-15
added Doxygen symbols for fontification

edam:
Added Doxygen symbols for fontification.


immerrr:
This looks useful, but I have a concern.

Luadoc keywords were highlighted, because they are specific to Lua ecosystem. Doxygen comments however are ubiquitous, and looks like packages exist for it already. I'd rather not have a copy of a subset of their functionality, and instead focus on ensuring that they operate fine with lua-mode's fontification.

Have you tried `doxymacs` or `highlight-doxygen`?

PR #173, opened 2020-09-04
Require rx at compile time

rnkn:
Given rx is a library of macros, it needs to be evaluated at compile time, otherwise many rx macros will be left undefined when installed via package-install.

n.b. Using eval-when-compile here will result in some rx functions being undefined at runtime.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -97,7 +97,8 @@
 
 (require 'comint)
 (require 'newcomment)
-(require 'rx)
+(eval-and-compile
+  (require 'rx))
 
 
 ;; rx-wrappers for Lua


npostavs:
I doubt this actually makes a difference, the byte-compiler already checks for `require`d libraries.

[(elisp) Eval During Compile](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html)

> Effectively ‘require’ is automatically ‘eval-and-compile’

rnkn:
@npostavs then it may surprise you to test this theory.

rnkn:
Sorry folks, I can no longer reproduce the issue I was getting. My mistake.

PR #179, opened 2020-11-05
Code cleanup/fix fixmes

nbfalcon:


lua-mode.el
```diff
@@ -2021,10 +2020,13 @@ Create a Lua process if one doesn't already exist."
...
 (defun lua-funcname-at-point ()
   "Get current Name { '.' Name } sequence."
-  ;; FIXME: copying/modifying syntax table for each call may incur a penalty
-  (with-syntax-table (copy-syntax-table)
-    (modify-syntax-entry ?. "_")
-    (current-word t)))
+  (save-excursion
+    (save-match-data
+      (re-search-backward "\\`\\|[^A-Za-z.]")
+      ;; NOTE: `point' will be either at the start of the buffer or on a
```
immerrr, line 2026: Could you add some tests for this to ensure the behaviour does not change? If I read this correctly, this new code will search for symbols all the way to the beginning of buffer, whereas `(current-word t)` only returned non-nil if point was within or adjacent to a word.

immerrr (changes requested):
Hello!

Thank you for submitting a PR with some code cleanups, it is much appreciated.

lua-mode.el
nbfalcon: No it won't. It will only handle the case of the symbol starting at the beginning of the buffer correctly. So the following will work:
```lua
|table|.insert({}, 5)
```
`(lua-funcname-at-point)`

| = possible cursor positions

nbfalcon: It handles all the cases of the previous function, but is more lenient about the cursor not being on a symbol already.
This will work:
```lua
|
-- 
table.insert({}, 5)
```

It is also removes trailing dots:
```lua
|
table.insert.({}, 5)
```

nbfalcon: My new function searches forward if a symbol isn't under the cursor. If that is an issue for you, I can fix it.

nbfalcon: Now it as strict as the old version: it only works if the cursor is really on an identifier. Note that trailing dots are still not in the output.

nbfalcon:
Now there is a comprehensive test-suite for `lua-funcname-at-point`.

lua-mode.el
immerrr: Leaving trailing dots out looks like a good addition, thanks!

The new code is now even more strict as the old one, it seems to change the behaviour for `foobar|()` case (with `|` being point): the old code returned `"foobar"` whereas the new one returns `nil`.

nbfalcon: I can fix that.

immerrr (approved):
LGTM

Thanks!

PR #180, opened 2020-11-05
`lua-start-proces': don't hang if already running

nbfalcon:
Due to waiting for output, that function hangs if a lua process is already
running, because `make-comint' doesn't recreate the process if it is already
running.

### DIFF
--- a/test/test-process.el
+++ b/test/test-process.el
@@ -1,0 +1,21 @@
+;;; test-funcname-at-point.el --- Test the repl functions
+
+;;; Commentary:
+
+;; Test functions to interact with the REPL.
+
+;;; Code:
+
+(describe "`lua-start-process'"
+  (it "doesn't hang for an already-running process"
+    ;; Acquire a *lua* repl buffer
+    (lua-start-process)
+    ;; Input some text
+    (with-current-buffer lua-process-buffer
+      (insert "table.insert"))
+    ;; Try restarting the repl buffer
+    (lua-start-process)
+    ;; `lua-start-process' shouldn't hang
+    (expect t)))
+
+;;; test-process.el ends here
--- a/test/test-process.el
+++ b/test/test-process.el
@@ -9,13 +9,16 @@
 (describe "`lua-start-process'"
   (it "doesn't hang for an already-running process"
     ;; Acquire a *lua* repl buffer
-    (lua-start-process)
-    ;; Input some text
-    (with-current-buffer lua-process-buffer
-      (insert "table.insert"))
-    ;; Try restarting the repl buffer
-    (lua-start-process)
-    ;; `lua-start-process' shouldn't hang
-    (expect t)))
+    (save-current-buffer
+      (funcall-interactively #'lua-start-process)
+      ;; Input some text
+      (with-current-buffer lua-process-buffer
+        (insert "table.insert"))
+      (switch-to-buffer (get-buffer-create "*scratch*"))
+      ;; Try restarting the repl buffer
+      (funcall-interactively #'lua-start-process)
+
+      ;; `lua-start-process' shouldn't hang, and it should have switched back.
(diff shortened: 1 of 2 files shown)


immerrr:
Hello!

Thank you for taking the time to submit a PR. If I am reading this correctly, it is to fix a bug. Could you provide a repro for that bug starting from `emacs -Q -l lua-mode.el`?

nbfalcon:
Sure.
1. `M-x lua-start-process`
2. Type in anything, but don't press RET (causing `lua-prompt-line` to yield nil)
2. Switch back to the lua source buffer
3. `M-x lua-start-process`now causes Emacs to hang, requiring C-g

nbfalcon:
Given that it is an alias for `run-lua`, it is by convention a function that is to open a lua repl, for interactive use.

immerrr:
Ok, so if I understand this correctly, the problem is that when `name` is non-unique, `make-comint` will not create a new buffer and thus will not create a new process, and then `lua-mode` tries to reinitialise the repl buffer and fails if there is data entered by the user.

I am slightly concerned about breaking backward compatibility in the following case:
- `(lua-start-process)`
- according to current implementation a new process for `lua -i` is started in `*lua*` buffer, `lua-process-buffer` points to it
- `(lua-start-process "my-new-process")`
- according to current implementation a new process for `lua -i` is started in `*my-new-process*` buffer, `lua-process-buffer` is switched to point to that, but `*lua*` buffer still exists

It seems like the situation could be fixed while maintaining backward compatibility for `lua-start-process` if it could avoid re-initialisation, e.g. by using the `lua--repl-buffer-p` variable. What do you think?

immerrr:
Great! Could you add a regression test to ensure the faulty behaviour is not introduced again?

nbfalcon:
How would I test if Emacs hangs?

immerrr:
You cannot test if something hangs (fun fact: it is called [Halting Problem](https://en.wikipedia.org/wiki/Halting_problem) and it is proven that there's no algorithm for it).

You can test the behaviour that caused problems in the past, then if you run the test and it hangs, you know there is something wrong with the program again.

nbfalcon:
Does buttercup have some sort of built-in support to check if some command didn't complete within n-seconds, should I come up with some idle-timer-based solution, or do you think it is good enough to have a hanging test-suite in that case?

nbfalcon:
There is now a regression test that hangs the test suite if the error reappears.

test/test-process.el
```diff
@@ -0,0 +1,21 @@
...
+  (it "doesn't hang for an already-running process"
+    ;; Acquire a *lua* repl buffer
+    (lua-start-process)
+    ;; Input some text
+    (with-current-buffer lua-process-buffer
+      (insert "table.insert"))
+    ;; Try restarting the repl buffer
+    (lua-start-process)
+    ;; `lua-start-process' shouldn't hang
+    (expect t)))
```
immerrr, line 19: There's `test-inferior-process.el` already, so this should probably go there.

Aside from that, couple of minor improvements
- do `(call-interactively 'lua-start-process)`, to mimick `M-x lua-start-process` because there's actually a piece of code that behaves differently depending on whether or not it is an interactive call
- instead of `expect t` it could check that `lua-start-process` switches to repl buffer even if it already exists

nbfalcon: Done.

nbfalcon: Also, you should use the sharp-quote (#') syntax when referring to function symbols, because doing that will give you byte-compiler warnings if the function is not defined:
```emacs-lisp
(defun foo ())

(defun bar () (funcall 'fo) ; no warning
  )
(defun bar2 () (funcall #'fo) ; byte-compiler warning here
  )
```

immerrr: Great, thanks!

immerrr (approved):

PR #199, opened 2022-01-04
Upgrade license to GPLv3

skangas:
Please consider upgrading the license to GPLv3.

Reasons why:
https://lists.gnu.org/archive/html/emacs-devel/2017-07/msg01069.html
https://www.gnu.org/licenses/rms-why-gplv3.html

The main file of this project says:

```
;; This program is free software; you can redistribute it and/or modify
;; it under the terms of the GNU General Public License as published by
;; the Free Software Foundation; either version 2 of the License, or
;; (at your option) any later version.
```

So this upgrade should be unproblematic.

Thanks!

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,281 +1,622 @@
                     GNU GENERAL PUBLIC LICENSE
-                       Version 2, June 1991
+                       Version 3, 29 June 2007
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
                             Preamble
 
-  The licenses for most software are designed to take away your
-freedom to share and change it.  By contrast, the GNU General Public
-License is intended to guarantee your freedom to share and change free
-software--to make sure the software is free for all its users.  This
-General Public License applies to most of the Free Software
-Foundation's software and to any other program whose authors commit to
-using it.  (Some other Free Software Foundation software is covered by
-the GNU Lesser General Public License instead.)  You can apply it to
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
 your programs, too.
 
   When we speak of free software, we are referring to freedom, not
 price.  Our General Public Licenses are designed to make sure that you
 have the freedom to distribute copies of free software (and charge for
-this service if you wish), that you receive source code or can get it
-if you want it, that you can change the software or use pieces of it
-in new free programs; and that you know you can do these things.
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
 
-  To protect your rights, we need to make restrictions that forbid
-anyone to deny you these rights or to ask you to surrender the rights.
-These restrictions translate to certain responsibilities for you if you
-distribute copies of the software, or if you modify it.
+  To protect your rights, we need to prevent others from denying you
+these rights or asking you to surrender the rights.  Therefore, you have
+certain responsibilities if you distribute copies of the software, or if
+you modify it: responsibilities to respect the freedom of others.
 
   For example, if you distribute copies of such a program, whether
-gratis or for a fee, you must give the recipients all the rights that
-you have.  You must make sure that they, too, receive or can get the
-source code.  And you must show them these terms so they know their
-rights.
+gratis or for a fee, you must pass on to the recipients the same
+freedoms that you received.  You must make sure that they, too, receive
+or can get the source code.  And you must show them these terms so they
+know their rights.
 
-  We protect your rights with two steps: (1) copyright the software, and
-(2) offer you this license which gives you legal permission to copy,
-distribute and/or modify the software.
+  Developers that use the GNU GPL protect your rights with two steps:
+(1) assert copyright on the software, and (2) offer you this License
+giving you legal permission to copy, distribute and/or modify it.
 
-  Also, for each author's protection and ours, we want to make certain
-that everyone understands that there is no warranty for this free
-software.  If the software is modified by someone else and passed on, we
-want its recipients to know that what they have is not the original, so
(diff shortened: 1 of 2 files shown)


immerrr:
I'm not exactly sure I'm able to relicense contributions by other contributors that were unequivocally made under GPLv2. 

Looks like the Internet agrees that without a CLA in place, it would require consent from all contributors [as discussed here on stackexchange](https://opensource.stackexchange.com/questions/1021/how-do-i-upgrade-from-gplv2-to-gplv3), [as explained when musescore upgraded from gplv2 to gplv3](https://musescore.org/en/node/320043) and [judging by the contributor tracking effort KDE project had to undertake to upgrade](https://community.kde.org/Guidelines_and_HOWTOs/Relicensing/KDE_Relicensing).

Do you have some other information backing your suggestion that the upgrade should be unproblematic?

skangas:
Yes, see for example:
https://www.gnu.org/licenses/gpl-faq.html#VersionThreeOrLater
(and the question after that)

The stackexchange discussion and the KDE case does not discuss the situation we are talking about here, but the musescore case is directly relevant. The reason they had to do all that is because their wording did not include the default "version 2 of the license, or (at your option) any later version". Instead, theirs simply said:

```
//  This program is free software; you can redistribute it and/or modify
//  it under the terms of the GNU General Public License version 2.

From: https://github.com/musescore/MuseScore/blob/5f3288511d1b75137ba4b709cb20e53bec3bba6b/src/main/main.cpp#L5-L8
```

Luckily, that's not the case here, as we _do_ have the "any later option" clause. Compare the above text to the one we have in lua-mode.el.

Hope that helps!

immerrr:
> Yes, see for example:
> https://www.gnu.org/licenses/gpl-faq.html#VersionThreeOrLater

Please, excuse my insistence, but I don't quite see yet how the answer you linked addresses relicensing from (GPLv2+ to GPLv3+).

I mean, yes, GPLv2+ would sidestep the problem that musescore and KDE had faced by enabling building on top of GPLv3 dependencies. But relicensing from GPLv2+ to GPLv3+ in itself is different as it could (would) take away some freedoms from the end users. To put it bluntly, if someone contributed to lua-mode with some intention to TIVOise it later on, they could be in their rights to object to the license change, as they would not be legally able to use derivatives of their own code afterwards. And I would agree with https://www.gnu.org/licenses/gpl-faq.html#OnlyLatestVersion, that it would be  

> a bad way to treat people.

even though the situation is a bit contrived.

Moreover, it seems to me there is even less reason to upgrade the license, because, as your link points out, if someone wanted to use lua-mode as a derivative work under GPLv3 they would be able to do so with the current license as is.

Am I not seeing something here?

immerrr:
And I'd like to reiterate my apologies. I just don't want to take rash decisions as license issues are complex and may be tough to untangle once a mistake has been made.

skangas:
I admit that I'm not 100 percent prepared for the question, as I never had to explain this situation in great detail. I first learned about it when the GPLv3 was released (and in the discussion leading up to it) when there was a lot of discussion about the words "version 2 of the License, or (at your option) any later version". Specifically that some projects had chosen not to use them and therefore couldn't be trivially upgraded. Since then, I have seen countless projects upgrade to GPLv3, and of course the vast majority of them with no copyright assignment or CLA's.

In other words, I am quite sure that upgrading is 100 % unproblematic, but I struggle to convince you.

I believe that due to this particular wording, few sources discuss the particular question you have. It might be seen as "too obvious", if you excuse the expression. For example, this FAQ just explains this regarding how to upgrade:
https://www.gnu.org/licenses/gpl-faq.html#v3HowToUpgrade

However, I realized that this is explicitly explained in the GPLv2:

>  Each version is given a distinguishing version number. If the Program specifies a version number of this License which applies to it and "any later version", you have the option of following the terms and conditions either of that version or of any later version published by the Free Software Foundation. If the Program does not specify a version number of this License, you may choose any version ever published by the Free Software Foundation. 

And in the GPLv3:

> If the Program specifies that a certain numbered version of the GNU General Public License “or any later version” applies to it, you have the option of following the terms and conditions either of that numbered version or of any later version published by the Free Software Foundation. If the Program does not specify a version number of the GNU General Public License, you may choose any version ever published by the Free Software Foundation.

As for your questions:

> To put it bluntly, if someone contributed to lua-mode with some intention to TIVOise it later on, they could be in their rights to object to the license change, as they would not be legally able to use derivatives of their own code afterwards.

The thing is, once you have granted some rights under the GPLv2 and distributed that work, you can never take those rights back. In other words, even if you protest, users are free to disregard your protest as they are granted the rights given to them by the GPL. This is a powerful thing about that license, which ensures software is free software going forward. But sure, if there are any "pro-tivoization" developers they might not enjoy the wording of the GPLv3.

> Moreover, it seems to me there is even less reason to upgrade the license, because, as your link points out, if someone wanted to use lua-mode as a derivative work under GPLv3 they would be able to do so with the current license as is.

Yes, but you yourself are also a user and have the same right. I think upgrading is just the least confusing alternative.

Please let me know if any of the above helped. Thanks for your patience.

immerrr:
> The thing is, once you have granted some rights under the GPLv2 and distributed that work, you can never take those rights back. In other words, even if you protest, users are free to disregard your protest as they are granted the rights given to them by the GPL.

Yeah, that is where it seems a bit tricky to me. This reads like it focuses on the rights of a subset of people (other users), but kinda forgets about the rights of the contributor. In my understanding that contributor still has their "author rights" to the code they contributed. The code they contributed, unless stated otherwise, came under the same license as the rest of the project, GPLv2+, and unless they agree to license it to me or the project under a different license, we may be in a bit of a pickle. I shouldn't be able to take their code and change the licensing in my (or someone else's) favour. To provide a similar blunt analogy, it is like when one puts a library as GPL, collects some contributions, and then willy-nilly changes the license to proprietary, this should not be legal.

In any case, I don't want you to spend more time on this, since this kind of details may not be your cup of tea. I'll sort this question of mine out with licensing@gnu.org, but happy to CC you in the conversation if you like.

immerrr:
Oh, and re

> Yes, but you yourself are also a user and have the same right. I think upgrading is just the least confusing alternative.

I agree, just like anyone, can _use_ it under GPLv2 (or GPLv3 or GPLv4 ...). The difference I see is that I cannot change the license of someone else's contribution so that a future user in the future can NOT use it under GPLv2 (thus taking away some of their rights). I would be able to do that if it was all my own code, or if the contributors signed a CLA that stipulated that. That's my understanding.

Again, apologies, don't want to bore you to death here or to stall a nice initiative :) Let me clarify this with GNU Licensing: if they are OK with it, I am too.

skangas:
Thanks for your patient approach, much appreciated.

Feel free to CC me into the conversation with licensing@gnu.org, I would be very interested to read what they have to say. :-)

immerrr:
> Feel free to CC me into the conversation with licensing@gnu.org, I would be very interested to read what they have to say.

Done. Hope you don't mind me getting your email from a mailing list.

immerrr:
This went off the radar for a while, let's get this in and sort out the complaints (if any) as they come.

immerrr:
Thanks

PR #202, opened 2022-08-01
Flymake

### LINKED ISSUE #201: Support luacheck via Flymake
Would you like to include a Flymake backend using luacheck in lua-mode? Then I'll make a PR. Otherwise I'll make it available as a standalone package on ELPA, but I guess it would be nicer to have this out-of-the-box.

astoff:
Add Flymake support, using the de facto standard flycheck program.

(I added a second commit the this PR adding the NonGNU ELPA badge to the Readme, and a third one bumping the version number so the new feature is immediately available on ELPA. But these are of course not essential and I can remove them.)

Closes #201.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Build Status](https://travis-ci.org/immerrr/lua-mode.svg?branch=master)](https://travis-ci.org/immerrr/lua-mode)
 [![MELPA](https://melpa.org/packages/lua-mode-badge.svg)](https://melpa.org/#/lua-mode)
 [![MELPA Stable](https://stable.melpa.org/packages/lua-mode-badge.svg)](https://stable.melpa.org/#/lua-mode)
+[![NonGNU ELPA](https://elpa.nongnu.org/nongnu/lua-mode.svg)](https://elpa.nongnu.org/nongnu/lua-mode.html)
 
 **lua-mode** is a major mode for editing Lua sources in Emacs.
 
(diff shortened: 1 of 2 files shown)


immerrr:
LGTM in general, nicely done :+1:

I have just noticed that this adds a `flymake-diagnostic-functions` by default. Will this play nicely with https://github.com/dmcooke/flymake-luacheck if some lua-mode user already has it installed?

astoff:
> Will this play nicely with https://github.com/dmcooke/flymake-luacheck if some lua-mode user already has it installed?

Flymake will run both backends in parallel, which is not desirable but not an error per se. My suggestion here, since the other package essentially depends on lua-mode but not vice versa, would be to have flymake-luacheck's setup function remove our backend from the list of Flymake backends. (It would not make as much sense to have the opposite logic, i.e., have the lua-mode package deal with an add-on package.)

immerrr:
> It would not make as much sense to have the opposite logic, i.e., have the lua-mode package deal with an add-on package.

I think there is at least a case to consider this logic.

With lua-mode first and foremost I try not to break things. There is not too much fancy new development out there that would fit the scope of this project that would not look & feel better as a separate package with its separate lifecycle. Flymake being another example of old and boring technology with a very limited scope is likely the reason I am for adding its backend to the core module.

Back to flymake-luacheck, even if we go and add a check for lua-mode's backend into it, there may still be users out there that have updated lua-mode but haven't updated flymake-luacheck. They would be affected by the double backend issue and would have to go edit their configs to disable one of the two. In this regard, lua-mode checking for flymake-luacheck's would simply recognise the fact that the latter became available ~2 years prior.

But now that I think about it, it may not work anyway. IIRC lua-mode hooks are only be run after the mode itself is initialised, so lua-mode wouldn't see the other backend at the moment of initialisation.

astoff:
> But now that I think about it, it may not work anyway. IIRC lua-mode hooks are only be run after the mode itself is initialised, so lua-mode wouldn't see the other backend at the moment of initialisation.

Good point, I didn't think about that. So I'd say there's nothing we can do about this here, really.

In any case, the breakage is very mild, and will not happen if the user updates both lua-mode and flymake-luacheck (assuming the latter will see an update to reflect the new feature of lua-mode).

astoff:
BTW, I've updated my pull request to use the `--ranges` feature of luacheck, which gives a more precise diagnostic location.

immerrr:
Merged, thank you!

PR #203, opened 2022-10-27
Enable lexical-binding for tests as required by buttercup

hillu:
Tests broke using buttercup 1.26, this commit fixes the breakage by simply enabling lexical binding in all test-related files.

See also: https://github.com/jorgenschaefer/emacs-buttercup/issues/226


immerrr:
LGTM, Thanks!

PR #207, opened 2023-06-13
No longer quote a lambda

tarsius:
Emacs v30's byte-compiler started to warn:

```
lib/lua-mode/lua-mode.el:333:10: Warning: (lambda (sym) ...)
quoted with ' rather than with #'
```

There is no need to quote at all.


immerrr:
Agreed, thanks!

PR #208, opened 2023-08-07
Use advice-add instead of obsolete defadvice

tarsius:
"New" advice has been added a long time ago and in Emacs 30 the use of `defadvice` finally results in a warning.


immerrr:
Thanks! `advice-add` apparently was introduced in 24.4, which is roughly the watermark I keep in mind for Emacs compatibility.

PR #211, opened 2023-10-22
fixed unescaped single-quotes

paralogismos:
Installing a fresh version of `lua-mode` on Emacs 29.1 I got several warnings such as:

```none
In lua-find-regexp:
lua-mode.el:994:2: Warning: docstring has wrong usage of unescaped single
    quotes (use \= or different quoting)
```

It seems that the byte compiler started [warning about some quoting mistakes in docstrings starting with Emacs 29](https://github.com/emacs-mirror/emacs/blob/6f87ee0e3a0727e16079778a4264e6e35cd7f3a8/etc/NEWS.29#L3637), and single-quotes now need to be escaped with `\=`. Apparently the unescaped single-quotes can be rendered incorrectly.

I escaped the single-quotes, except for some which were paired. I changed those to backquote/single-quote pairs like `` `foo' ``; it seems that these pairs are fine.

The only byte-compiler warnings I saw after were for three docstrings wider than 80 characters, and I left those alone.

### DIFF
--- a/lua-mode.el
+++ b/lua-mode.el
@@ -994,7 +994,7 @@ placed before the string."
 (defun lua-find-regexp (direction regexp &optional limit)
   "Searches for a regular expression in the direction specified.
 
-Direction is one of 'forward and 'backward.
+Direction is one of \\='forward and \\='backward.
 
 Matches in comments and strings are ignored. If the regexp is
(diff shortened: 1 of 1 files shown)


immerrr:
Looks good, thanks!

PR #214, opened 2024-04-16
Use eval in lexical scope to fix tests

manphiz:
* Buttercup since 1.34 starts to support oclosure, so eval will be used in dynamic scope by default.
* This patch ensures that eval is using lexical scope to retain previous behavior and fix tests in buttercup >=1.34.


manphiz:
Closing this in favor of https://github.com/immerrr/lua-mode/pull/215

PR #219, opened 2025-11-05
Silence Emacs 31.1 byte-compiler

tarsius:
```
lua-mode.el:733:27: Warning: ‘hs-special-modes-alist’ is an
obsolete variable (as of 31.1); use ‘hs-modes-alist’ instead.
```


tarsius:
I've noticed that this is now maintained as part of Emacs, so I am closing this. (Please update the readme accordingly.)

